PR #94, opened 2020-07-15
Add: [JP] add ocr (using tesseract command)

ferina8-14:
1. Add "OcrJapanese" class to provide Japanese ocr
2. Add logics for jp commissions
3. update requirements.txt
4. Fix: recheck all opponents after updating them (exercise)

I thought using tesseract command would be better than training cnocr for Japanese letters.
It seems that it can almost recognize letters in commission names except for Roman numerals.


LmeSzinc:
If you can provide me the fonts used in JP and a txt file of japanese charset, i can train an OCR model in japanese
and i want to know what's the OCR result of those Roman numerals in tesseract, maybe you can strip them

The reason I choose cnocr instead of tesseract or online ocr api
1. Tesseract has low accuacy in chinese, there are 10k to 20k modern chinese characters, but only 2k to 3k is often used.
    Tesseract often give result of those rarely use, it's true that they looks very similar.
2. cnocr is easy to use, users don't need to do anything, just pip install cnocr
    Don't need other installation, don't need to register and get api token

ferina8-14:
I'm sorry to be late.

As for using tesseract,  I understand that using cnocr is better if we can train the model for JP.  

About the fonts, I'm not sure what fonts they are using but 'HGP創英ﾌﾟﾚｾﾞﾝｽEB' looks similar.
I will send the ttf file to you.
Also, I can send you the images of commission names (not all).

And Roman numerals are recognized like this.
"Ⅰ" => ["I", "1"]
"Ⅲ" => ["川", "利"]

Thank you.

[HGPSoeiPresenceEB-02.zip](https://github.com/LmeSzinc/AzurLaneAutoScript/files/4948593/HGPSoeiPresenceEB-02.zip)
[img_jp_commission.zip](https://github.com/LmeSzinc/AzurLaneAutoScript/files/4948594/img_jp_commission.zip)

LmeSzinc:
I think "HGP創英ﾌﾟﾚｾﾞﾝｽEB" is not the one. You can google some website to find/idendity the fonts.

fonts
Take 00001.png as an example, you can download this image and zoom in.
It's a sans serif font, but "HGP創英ﾌﾟﾚｾﾞﾝｽEB" is a serif font. (Google "serif vs sans serif" to know the details)
The reset of the fonts on the image were found in [fontmgr.lua](https://github.com/Dimbreath/AzurLaneData/blob/master/ja-JP/mgr/fontmgr.lua), I tried the fonts I understand.
Take a closer look at letter "演", they are different.
If still cannot find the font, you can provide a group of commonly used Japanese fonts

About the charset, I found [EasyOCR](https://github.com/JaidedAI/EasyOCR/blob/master/easyocr/character/ja_char.txt). It should cover commonly used japanese letters. 
Hundreds of samples is not enough. Current model was trained on 500K samples, and the models in cnocr were trained on 1M samples (they said). Training dataset is auto-generated, using [text_renderer](https://github.com/breezedeus/text_renderer), don't need to collect the image in game.

PR #97, opened 2020-07-17
Updated droprates module

nicholasjin:
Module now stores an archive of previous items, and users no longer need to copy the item_template directory around. Updated documentation.


LmeSzinc:
Closing this PR, because the item_statistics has been rewritten to fit new OCR model.
Also, item_statistics.py has been split into several files, in order to do real-time statistics in the future, such as auto buy items in shop.

PR #155, opened 2020-08-12
Fix: Enhance Timing Issue

nEEtdo0d:
Not confident will actually fix the issue altogether but this is how the original implementation had it, I had hoped to reduce some time with what I had. The primary issue I believe is timing of the screenshot.


nEEtdo0d:
Updated filter assets for generalized usage, original assets success varied from user to user
Small enhancements and fixes, core implementation unchanged however now returns boolean to allow verification to callee methods.

PR #173, opened 2020-09-03
Fix: Handle maintenance announcement at login

noname94:
03-09-2020_03-10-05_AM


whoamikyo (changes requested):
I don't think it's necessary to force the configuration only for the JP server, at least on the EN server appears to be the same.

whoamikyo (approved):
Thanks.

PR #174, opened 2020-09-07
Add: [JP] Research support

noname94:
JP version doesn't have any research code like "D-722-MI" at page_research, so we have to enter detail pages for more information.
research


whoamikyo (commented):
Too bad JP version doesn't use same name, However, I think that its implementation differs somewhat from the pre-established concept, I think the correct way to implement the Research module in JP version would be:

- train an OCR model in japanese
- use the dev_tools research_extractor to extract research names in japanese from https://github.com/Dimbreath/AzurLaneData/blob/master/ja-JP/sharecfg/technology_data_template.lua

there has already been a discussion about train OCR here https://github.com/LmeSzinc/AzurLaneAutoScript/pull/94

noname94:
> 
> 
> Too bad JP version doesn't use same name, However, I think that its implementation differs somewhat from the pre-established concept, I think the correct way to implement the Research module in JP version would be:
> 
>     * train an OCR model in japanese
> 
>     * use the dev_tools research_extractor to extract research names in japanese from https://github.com/Dimbreath/AzurLaneData/blob/master/ja-JP/sharecfg/technology_data_template.lua
> 
> 
> there has already been a discussion about train OCR here #94

The problem is that the word replaced the code is just a copy of the research category. For example, “艤装解析” means "Rigging analysis".
As a result, there is no enough information at page_research. The only way is to enter the detail pages.

LmeSzinc:
noname94 has discussed with me before making a PR. It's true that page_research in JP contains too few infomation.

LmeSzinc:
I also imagined detecting the reward items and project color in page_research, but still cannot distinguish some projects, such as `T1` and `T2`, `D8` and `D0.5`. Checking projects in detailed sounds dumb, but seems to be the easy way.

PR #189, opened 2020-09-20
CN, EN, JP: Add CampaignAB support for Chapter T and HT

nEEtdo0d:


nEEtdo0d:
Ran a trial run, no outstanding errors I think I caught everything needed to support for servers

whoamikyo (approved):
I just don't see any reason to add support for HT right now, since hard mode has no bonus at the moment and we don't even know if they are going to launch other maps with that same nomenclature.

PR #193, opened 2020-09-22
Add: predict_submarine

noname94:
The ammo icon of submarines could make 'is_fleet' labeled to wrong grids. The main popurse of the commit is to fix this.


noname94:
is_submarine should not be merged because we wipe_out 'is_xxx' flags after the movement of a fleet, but submarines take their movement after that. As a result, there can be 2 grids flagged as 'is_submarine' at the same time.
On the other hand, it is meaningless to get the exact position of our submarines, so it is OK to ignore it.

noname94:
If is_submarine is merged, we'll need to delete it every turn, otherwise multiple grids will be flagged "ss". Since it makes nonsense to know the exact position of our submarines, and the main purpose is to prevent "Fl" wrongly flagged, I just leave is_submarine unmerged.

PR #202, opened 2020-09-25
fix  ht2 do not switch fleet to clear boss

ashesofdream:


LmeSzinc:
Reverted

- Please don't use fleet_boss when boss_spawn <= 5
- Please don't make PR for personal battle logic

PR #205, opened 2020-09-26
Fix: fix the stable check before emergency repairing may not work on …

noname94:
…potato pcs


module/combat/combat.py
```diff
@@ -166,8 +166,10 @@ def handle_emergency_repair_use(self):
         if self.appear(BATTLE_PREPARATION) and self.appear(EMERGENCY_REPAIR_AVAILABLE):
             # When entering battle_preparation page (or after emergency repairing), the emergency icon is active by default, 
             # even if nothing to use. After a short animation, everything shows as usual. Using fleet power number as a 
-            # stable checker.
-            self.wait_until_stable(PREPARATION_STABLE_CHECKER)
+            # stable checker. First wait for it to be non-zero, then wait for it to be stable.
+            self.wait_until_disappear(MAIN_FLEET_POWER_ZERO, offset=(20, 20))
+            stable_checker = Button(area=MAIN_FLEET_POWER_ZERO.area, color=(), button=MAIN_FLEET_POWER_ZERO.button, name='')
```
LmeSzinc, line 171: It's recommended to use `name='STABLE_CHECKER'`. If name is not set, it will traceback code to set name.
This a problem left by history (when Alas needs to type coordinates manully)

noname94: Got it

PR #206, opened 2020-09-27
Opt: reward receiving logic optimization

noname94:
To ensure that nothing is pending before exiting page_reward. It also avoids using research_receive() function in RewardResearch class which might be unstable because of too many sleep().

### DIFF
--- a/module/research/research.py
+++ b/module/research/research.py
@@ -341,11 +341,12 @@ class RewardResearch(ResearchSelector):
         if not self.config.ENABLE_RESEARCH_REWARD:
             return False
 
-        if not self.appear(RESEARCH_FINISHED) and not self.appear(RESEARCH_PENDING, offset=(20, 20)):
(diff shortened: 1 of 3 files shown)


module/research/research.py
```diff
@@ -341,11 +341,12 @@ def handle_research_reward(self):
         if not self.config.ENABLE_RESEARCH_REWARD:
             return False
 
-        if not self.appear(RESEARCH_FINISHED) and not self.appear(RESEARCH_PENDING, offset=(20, 20)):
-            logger.info('No research finished or pending')
+        if not self.appear(RESEARCH_PENDING, offset=(20, 20)):
+            logger.info('No research pending')
```
LmeSzinc, line 345: When drop screenshot enabled, research receive is in page_research, because research stats need research name.
So here needs RESEARCH_FINISHED.

PR #212, opened 2020-09-29
Opt: normalize before calculation in hp balancer

noname94:
To avoid sth strange like: threshold = 0.2, hp = [0.05, 0.2, 0], no reposition.


LmeSzinc:
`hp = np.array(hp) / max(hp)` is not a good idea, things will be changed when ships are still in high HP.
Also, i think it should trigger a retreat in this situation, instead of continuing.

PR #219, opened 2020-10-08
Update: Adjust Dorm Reward to default collect, feed configurable. Ren…

nEEtdo0d:
…amed dorm_interval to feed_interval for feed record, added collect_interval for collect_record

Update: Enhance limit to 4 ship checks per category, no restart ship check number in-between method calls. Reduce trapped check handling limit from 6 to 4. Ternary for end_activate_timer based on DEVICE_CONTROL_METHOD

TODO: Adjust CN dictionary.py and argparser.py based on Dorm Reward update


LmeSzinc:
1. Clearify dorm_run() and handle_dorm()
dorm_run() is a function to do things in dorm. It shouldn't use config, and it can run outside of Alas.
handle_dorm() is a function to call dorm_run() according to user config, also handles game UI.
You can also find similar functions such as commission_start() and handle_commission_start()
So implement `dorm_run(feed=True, collect=True)`, and handle_dorm() calls dorm_run() according to ENABLE_DORM_REWARD and last record time.
2. Name prefix
Configs in template.ini, should start with `dorm_`, functions in DormReward should start with `dorm_`, configs in AzurLaneConfig should start with `DORM_`. So rename `collect_interval` to `dorm_collect_interval`, etc.
3. Separate this modification of dorm and enhancement into 2 commits. If possible, rebase to delete the first commit which contains two different things. Only leave 2 commits in this PR.

nEEtdo0d:
LmeSzinc I've updated the implementation as requested based on 1 and 2. I'm not sure how I might be able to do 3 with the commit being in the middle and performing a rebase from there.  If preferred, I could independently fork and make a new PR altogether that only contains the dorm enhancements.

Thanks

PR #241, opened 2020-11-13
Add: War Archives Automation

nEEtdo0d:
Supports all EN available campaign maps

Seems stable; one encounter of unable to story skip after entering the stage map. It was trying to click STRATEGY_OPEN but couldn't since the the story sequence overlay was blocking it.

Unsure whether CN/JP uses scrolling or turning page mechanism to display the other events beyond the 4 table grid.  Either way, an modded implementation of template searching would be needed to handle this.


campaign/campaign_war_archives/campaign_base.py
```diff
@@ -0,0 +1,123 @@
...
+            type(str): ex or sp
+        """
+        for n in range(3):
+            self.device.screenshot()
+            if type in 'ex':
+                self.appear_then_click(WAR_ARCHIVES_EX_OFF)
+            elif type in 'sp':
+                self.appear_then_click(WAR_ARCHIVES_SP_OFF)
+            else:
+                raise CampaignNameError
```
LmeSzinc, line 31: use Switch class with `is_selector=True`, like COMMISSION_SWITCH
```
COMMISSION_SWITCH = Switch('Commission_switch', is_selector=True)
COMMISSION_SWITCH.add_status('daily', COMMISSION_DAILY)
COMMISSION_SWITCH.add_status('urgent', COMMISSION_URGENT)
```

LmeSzinc: Don't use `type` as variable name, `type()` is a internal function of python

campaign/campaign_war_archives/campaign_base.py
```diff
@@ -0,0 +1,123 @@
...
+            self._set_archives_view(type)
+            self.handle_stage_icon_spawn()
+
+            # Acquire approriate event entrance based on template
+            # Wait for stage_icon_spawn
+            archives_entrance = self._get_archives_entrance(self.config.WAR_ARCHIVES_NAME)
+            self.ui_click(archives_entrance, appear_button=WAR_ARCHIVES_CHECK, check_button=WAR_ARCHIVES_CAMPAIGN_CHECK,
+                          skip_first_screenshot=True)
+            self.ensure_no_story()
+            self.handle_stage_icon_spawn()
```
LmeSzinc, line 97: `self.ensure_no_story()` may not works here, it will still stuck in `self.ui_click`
Alas can't handle story during page switching at the moment. Users need to skip it manully.

campaign/campaign_war_archives/campaign_base.py
```diff
@@ -0,0 +1,123 @@
...
+        event campaign map
+
+        Args:
+            type (str): 'ex' or 'sp'.
+        """
+        # First run, regardless of current location
+        # Start from page_main
+        if self.first_run:
+            self.first_run = False
+            self.ui_ensure(destination=page_main)
```
LmeSzinc, line 79: No need to goto page _main
If already in a war achieve, go back to page_achieves. If not, run as normal.

module/handler/info_handler.py
```diff
@@ -89,6 +91,16 @@ def handle_combat_low_emotion(self):
...
+    def handle_use_data_key(self):
+        if not self.appear(POPUP_CONFIRM) and not self.appear(POPUP_CANCEL):
+            return False
+
+        ocr_result = DATA_KEY_USE.ocr(self.device.image)
+        if 'Data Key' in ocr_result:
+            return self.handle_popup_confirm('DATA_KEY_USE')
+
+        return False
+
```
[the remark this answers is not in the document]
LmeSzinc: Add `USE_DATA_KEY = False`  in config class, set it to True when running war achieves.
In `handle_use_data_key()`, use this to skip the whole detection process when not running war achieves.
```
if not self.config.USE_DATA_KEY:
    return False
```

campaign/campaign_war_archives/campaign_base.py
```diff
@@ -0,0 +1,123 @@
...
+            if chapter in 'ab':
+                self.campaign_ensure_mode('normal')
+            elif chapter in 'cd':
+                self.campaign_ensure_mode('hard')
+            self.campaign_ensure_chapter(index=chapter)
+        elif chapter == 'sp':
+            self.ui_goto_archives_campaign('sp')
+            self.campaign_ensure_chapter(index=chapter)
+        else:
+            logger.warning(f'Unknown campaign chapter: {name}')
```
LmeSzinc, line 123: Override `ui_goto_event()` and `ui_goto_sp()`, because some event maps already override `campaign_set_chapter()` to handle abnormal stage orders.

LmeSzinc (commented):
1. Add a `IS_WAR_ACHIEVES` option in dev_tool/map_extractor.py to auto add `from ..campaign_war_archives.campaign_base import CampaignBase`.
2. Add only Visitors Dyed in Red at the moment,, other maps need simple tests. Some maps with chinese siren template names are not runable, `MAP_SIREN_TEMPLATE = ['sairenquzhu', 'sairenqingxun']`
Need to test if current template can match them, if not, make new templates.
3. Delete EX maps.

module/handler/info_handler.py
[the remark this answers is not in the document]
nEEtdo0d: I had switched OCR because the `self.appear` sometimes failed to evaluate to true but just noticed that uses `button.appear_on` rather than `button.match`, so I'll try using `offset=1` to see if that makes a difference.

campaign/campaign_war_archives/campaign_base.py
nEEtdo0d: I opted in retaining the 'first_run' behavior since ui_goto_event is also called after every map completion as well. Otherwise, on the 2nd or Xth run, time will be used to going back to page_archives and getting back to where it already was and data key only gives 15 minutes.  Of course, can't do much about reward or stop triggers those are just acceptable loss of time.

But if you think otherwise, I have no problem dropping it either.

campaign/campaign_war_archives/campaign_base.py
nEEtdo0d: Great idea! I hadn't thought about overriding those methods that makes way more sense!

module/config/argparser.py
```diff
@@ -77,6 +77,11 @@ def main(ini_name=''):
     event_latest = sorted([f for f in event_folder], reverse=True)[0]
     event_folder = [dic_ini_to_gui.get(f, f) for f in event_folder][::-1]
     event_latest = dic_ini_to_gui.get(event_latest, event_latest)
+    # Archives list
+    archives_folder = [f for f in dic_event[server.server].values() if f.startswith('war_archives_')]
+    archives_latest = sorted([f for f in archives_folder], reverse=True)[0]
+    archives_folder = [dic_ini_to_gui.get(f, f) for f in archives_folder][::-1]
+    archives_latest = dic_ini_to_gui.get(archives_latest, archives_latest)
```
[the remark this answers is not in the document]
nEEtdo0d: Since I haven't unlocked all the stages yet I did do practice runs on earlier stages that did have sirens. The siren template 'DD' and 'CL' for Winter's Crown and Fallen Wings worked. Divergent Chessboard doesn't have any sirens according to the wiki.

Do you still prefer I delete these EX maps?

module/handler/info_handler.py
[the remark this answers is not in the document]
LmeSzinc: You can use `self.appear()` with `offset=(20, 20)`, `offset=True` equals `offset=(0, 30)`

campaign/campaign_war_archives/campaign_base.py
LmeSzinc: Oh, i understand what `first_run` is.
About the reward loop, you can disable it when using data key, and check reward after the key ends.
```
backup = self.config.cover(ENABLE_REWARD=False)
# Do something
backup.recover()
```

module/config/argparser.py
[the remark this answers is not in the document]
LmeSzinc: Sorry, word `EX` is ambiguous. I mean maps like `campaign/war_archives_20181026_en/ex.py` and `campaign/war_archives_20191031_en/ex.py`

campaign/campaign_war_archives/campaign_base.py
nEEtdo0d: Good idea but having trouble where to put these lines in which files and where. If you were suggesting cover in one file and recover in another file but variable sharing would be difficult I think.

Alternatively, I was thinking overriding `handle_reward` like `triggered_stop_condition`
```
backup = ""
if self.in_archives_campaign():
    if not self.appear(CAMPAIGN_OPEN, offset=(20, 20)):
        backup = self.config.cover(ENABLE_REWARD=False)
result = super().handle_reward()
if backup != "":
    backup.recover()
return result
```
CAMPAIGN_OPEN would be a new asset of a button displayed below the data key counter, in EN says "Not open yet" and when data key used, it changes to a timer print out.

LmeSzinc: Using CAMPAIGN_OPEN woule be better. Since you override handle_reward, no need to cover attributes in config
```
def handle_reward(self):
    if not self.appear(CAMPAIGN_OPEN, offset=(20, 20)):
        return False
    return super().handle_reward()
```

LmeSzinc: Is it possible to make `is_in_campaign` faster?
With no screenshot taken inside the function.

nEEtdo0d: The best I can think of is an internal bool like `first_run`, `in_campaign = False` in war_archives campaign_base. This might be a way to avoid screenshot altogether. Also may not need `first_run` if using this instead and maybe more clean to use a public setter method rather than direct variable update. But I'd think still using screenshot is safest option to check whether in that page to ensure those whether those actions can be performed or not. In theory the bool should work as well, assuming only select actions will ever interact and update it. 

campaign_base.py
```
in_campaign = False
def ui_goto_archives_campaign(self, mode='ex'):
     if not self.in_campaign:
          # Change needed pages
          self.in_campaign = True
```
war_archives.py
```
def handle_reward():
     if self.campaign.in_campaign and not self.appear(CAMPAIGN_OPEN), offset=(20, 20)):
         return False
     if super().handle_reward():
         self.campaign.in_campaign = False
         return True
     return False
def triggered_stop_condition():
    if self.campaign.in_campaign:
        # Same stuff

Comment (LmeSzinc) [reply campaign/campaign_war_archives/campaign_base.py]: How about replace
```
if self.first_run or not self.in_archives_campaign():
    result = self.ui_ensure(destination=page_archives)
```
with:
```
if self.first_run or not self.appear(WAR_ARCHIVES_CAMPAIGN_CHECK, offset=(20, 20)):
    result = self.ui_ensure(destination=page_archives)
```

Comment (nEEtdo0d) [reply campaign/campaign_war_archives/campaign_base.py]: Oh right, a much simpler answer. This is what the helper method is technically doing just in a looping behavior to make sure it was in that screen before moving forward. My PC can be slow sometimes so thought it was better to give it some leeway for safety.

I think I see your point though, there is already a loop in effect via `ensure_campaign_ui()` so will inherently retry again until ensured or times out. Another loop will just slow things down.

I'll change to `self.appear(WAR_ARCHIVES_CAMPAIGN_CHECK, offset=(20, 20))` from what I have `self.in_archives_campaign()` in their respective files.

Comment (nEEtdo0d) [reply campaign/campaign_war_archives/campaign_base.py]:
```
    def handle_reward(self):
        if not self.appear(WAR_ARCHIVES_CAMPAIGN_CLOSED, offset=(20, 20)):
            logger.info('Delaying reward check until data key has been used up')
            return False
        return super().handle_reward()
```
Encountered Device has no attribute called image which makes sense since this is used in run almost immediately.  `hasattr()` check to mitigate this?

Comment (LmeSzinc) [reply campaign/campaign_war_archives/campaign_base.py]: Check if `Device` has attribute `image`, if not, take one screenshot.

Comment (nEEtdo0d) [reply campaign/campaign_war_archives/campaign_base.py]: That does the trick.

New issue, not sure how I should cut the image for that "Not yet open' button text.  The background is relatively transparent and so getting mixed results based on selected archives stage map.

[captures](https://imgur.com/a/Y9Z4NLI)

I've tried whole text and the empty space on the left since that is where the timer would be. A fraction of the text may vary in results like the 'N' or 'O' isn't that different if the appearance relies more on color. Would offset be better to use? Alternatively, maybe use OCR like how I did with the 'Data Key' before.

Comment (LmeSzinc) [reply campaign/campaign_war_archives/campaign_base.py]: You can see how alas detect ambush evade  success and ambush evade failed. They are both white letters on a semi transparent background.
However, I have another way to detect this. `Not yet open` is on the left, blank area on the right. While it's open, white letters on the right. You can calculate the maximum or average RGB on the right to do the trick.

Comment (nEEtdo0d) [reply campaign/campaign_war_archives/campaign_base.py]: Would like to clarify whether this is what you had in mind, I used `button.load_color()` and used image of the blank area
`Not yet open` RGB from 3 different images
[43.34810811 50.11027027 74.52756757]
[66.45675676 61.49837838 83.05945946]
[49.90972973 49.82108108 73.26486486]

`Open` RGB from 1 image
[90.05351351  96.7572973  111.13459459]

Use the `Open` RGB average as the expectation and compare against a live image's average? If lower then considered `Not yet open` else if higher then means `Open`? Perhaps some plus/minus leeway is needed to account for margin of error?

Average can be calculated like this, taken from [here](https://stackoverflow.com/questions/12703871/benchmarking-functions-for-computing-average-rgb-value-of-images)
`tuple(button.load_color().mean(axis=0))` which I think is numpy array

Also if in any other page, should probably safeguard this validation process with the same `WAR_ARCHIVES_CAMPAIGN_CHECK` condition

Comment (nEEtdo0d) [reply campaign/campaign_war_archives/campaign_base.py]: [Created Assets](https://imgur.com/a/8mciqpy)

Those are the assets I'm using for testing, surprisingly using `self.appear()` on the `N` does work for all the screen captures but of course that would mean every server would have to provide their own image while this RGB method is generic enough that no one has to make asset updates.

Comment (LmeSzinc) [reply campaign/campaign_war_archives/campaign_base.py]: WAR_ARCHIVES_TIMER
The info tab on top right is a 60% transparent black background. So average color of background will smaller than 255 * 40% = 102.
```
if np.mean(self.image_area(WAR_ARCHIVES_TIMER)) > 102:
    # Timer running
else:
    # Not yet open
```

nEEtdo0d: How did you come up with 60% transparency? Just a rough estimate? Great idea to box the asset as close as possible to the white text for the RGB to incorporate more white emphasis, those previous assets didn't quite make good RGB values for comparisons since it included too much of the other colors.

Ran through a Winter's Crown D3 sequence, no interruptions and handle_reward executed timely based on whether the timer was there or not. I think PR is ready unless I may have missed something. Let me know if I did.

Thanks

LmeSzinc: Just estimate by color.

This PR will be merged soon.

PR #246, opened 2020-11-20
Add: [JP] meowfficer assets added

noname94:


campaign/campaign_main/campaign_7_2.py
```diff
@@ -21,7 +21,7 @@
     30 30 30 30 30 30 30 30
 '''
 MAP.camera_data = ['D3', 'D2']
-MAP.camera_data_spawn_point = ['D3', 'D2']
+MAP.camera_data_spawn_point = ['A2']
```
LmeSzinc, line 24: `camera_data_spawn_point` is not used like this.
It's the camera position that must be scanned after entering map. If camera sight didn't covered the full map, there will be enemy missing.

Can you describe what's `unexpected combat` ?

LmeSzinc: When alas scanning the map, if all objects in spawn_data have been detected, full scan will be early stopped. camera_data_spawn_point is used to ensure that all fleet spawn points have been scanned, because fleets are not in spawn_data.

noname94: 21-11-2020_12-53-45_AM
The green arrow bumps and might cover the the scale icon when it is up (and goes normal when it is down). Here the enemy at A3 is missing in ALAS. Both the camera_data points D3 and D2 have this problem. If we scan at An, this won't happen because the arrow will be in the middle while the scale icon is on the top-left corner.
If we missed the enemy at A3 during both the two scan at D3 and D2, the boss fleet will step on A3, which is the unexpected combat.

LmeSzinc: Alas can handle this automaticly without any modification.
`CampaignBase.missing_predict()` in `module/map/map_base.py` is designed for such situation: to predict things being covered by ships/sirens/arrows. It can't get prediction everytime, but enough to work in 7-2.
If fleet_1 is on A5 and fleet_2 is on H1. There's only 1 enemy missing and 1 grid both covered and may_enemy, which is A3. Then A3 will be predicted to be '0E' (is_enemy=True, enemy_scale=0, enemy_genre='Enemy')

noname94: I see. Thanks for the explanation! Reverted.

PR #304, opened 2021-01-14
Fix: Assets, InfoGuild, and RewardGuild

nEEtdo0d:
PR includes:
- Remove unused asset, added accidently in another PR
- Fix filter for index ship 'others' type
- Add new get items from meowfficer dive during os, 2 row acquired items. Possible update in future if 3 row exists


nEEtdo0d:
Additional commits discovery after extended usage:
- Inheritance of InfoGuild into Combat collides with Equipment module for method _view_swipe, changed signature names
- Guild collect often exit early due to mis-clicks of target area either being too large or possibly laggy response, reduce area for reliable target clicks or at least shown to be. May return to extend the timer from 1.5s to 3+s

LmeSzinc:
I think we should reduce the amount of png files. [This commit](https://github.com/LmeSzinc/AzurLaneAutoScript/pull/304/commits/0ace1d1e8c38ac3bcfda5b43ebae510cc5ab6d7f) contains too many assets.

You can use some simple color tricks to detect guild mission. Example:
guild_mission
```
for image in images:
    az.device.image = image
    self = az
    GUILD_MISSION = GUILD_MISSION_ACCEPT_AXIS

    # Test code here
    r, g, b = get_color(self.device.image, GUILD_MISSION.area)
    if g > max(r, b) - 10:
        print('Guild mission finished this week')
    else:
        # unfinished mission range from about 240 to 322
        if self.image_color_count(GUILD_MISSION, color=(255, 255, 255), threshold=180, count=400):
            print('Guild mission button active')
        else:
            print('Guild mission unfinished')
```
output:
```
Guild mission unfinished
Guild mission unfinished
Guild mission button active
Guild mission finished this week
Guild mission unfinished
Guild mission button active
```

LmeSzinc:
So do guild supply. Example:
guild_supply

```
for image in images:
    az.device.image = image
    self = az
    GUILD_SUPPLY = GUILD_SUPPLY_REWARDS_AXIS

    # Test code here
    color = get_color(self.device.image, GUILD_SUPPLY.area)
    if np.max(color) > np.mean(color) + 25:
        print('Guild reward button active')
    else:
        print('Guild reward already collected')
```
output:
```
Guild reward already collected
Guild reward button active
Guild reward button active
```

nEEtdo0d:
The many assets might not have been the best idea since my goal was to minimize the click area.

My `InfoGuild` was for some reason getting stuck even with `ui_back`, had recorded having clicked the button but no pop up response and `logistics_collect` had ended early. It had somehow returned to the campaign loop check but it hadn't exited the `page_guild`. However using these `.BUTTON` assets for some reason made it more reliable after that.

But in the end lag might have been the actual culprit.  I imagine perhaps it did click, `POPUP_CONFIRM` appeared too late and exited `logistics_collect`, it did eventually show up and `ui_back` closed it and also exited believing it was outside of `page_guild` but the image might have been `POPUP_CONFIRM` in the midst of closing. Perhaps just increasing `confirm_timer` would have been sufficient. Though the odd thing is that a normal reward loop had never encountered a weird situation like that.

I will remove the assets since they were primarily for `InfoGuild`

LmeSzinc:
> Did you still want me to integrate these color tests and replace appear_then_click within logistics_collect if so? I assume I'd just use self.device.click approriately based on the condition.

Yes, and I want to delete `is_azur_affiliation`, make it working in both azur and axis. It really took me some days to collect screenshots in different affiliation, so I afraid it would be hard to maintain those assets in the future.

nEEtdo0d:
May need to retain `is_azur_affiliation` its used to determine when a while loop can end via `LOGISTICS_CHECK_AZUR/AXIS` and also by the exchange procedures to determine the letter color before executing `ocr`

Also I've noticed when going to logistics or any sidebar option. More often than not, takes a while for the page to fully load. The image captured last often is empty so a pseudo `wait_until_appear` delay is needed after clicking sidebar. I've been able to make due with this if trying to not depend on `is_azur_affiliation`. 
```
        confirm_timer = Timer(1.5, count=3).start()
        while 1:
            if skip_first_screenshot:
                skip_first_screenshot = False
            else:
                self.device.screenshot()

            # End
            if self.appear(GUILD_LOGISTICS_CHECK_AXIS) or self.appear(GUILD_LOGISTICS_CHECK_AZUR):
                if confirm_timer.reached():
                    return True
            else:
                confirm_timer.reset()
```

Would either of your snippets also work for `LOGISITCS_CHECK_AZUR/AXIS` and `OCR_GUILD_EXCHANGE_LIMIT`?

nEEtdo0d:
Through some trial and error just pretty much re-arranging your snippet, getting stable results using 
```
    def _guild_logistics_ensure(self):
        """
        Color sample the GUILD_LOGISTICS_ENSURE
        to determine whether is currently
        visible or not

        Pages:
            in: GUILD_LOGISTICS
            out: GUILD_LOGISTICS
        """
        # Axis (181, 97, 99) and Azur (148, 178, 255)
        if self.image_color_count(GUILD_LOGISTICS_CHECK_AXIS, color=(181, 97, 99), threshold=180, count=400) or \
                self.image_color_count(GUILD_LOGISTICS_CHECK_AXIS, color=(148, 178, 255), threshold=180, count=400):
            return True
        else:
            return False
```

For places like these
```
while 1:
....
    # End
    if self._guild_logistics_ensure():
        if not self.info_bar_count() and confirm_timer.reached():
            break
        else:
            confirm_timer.reset()
...
```
```
self.ui_click(target_button, check_button=POPUP_CONFIRM,
                   appear_button=self._guild_logistics_ensure, skip_first_screenshot=True)
```

nEEtdo0d:
Interesting scenario with the operations though, with what they added they prioritize newest events.
17-01-2021_03-16-07_PM
17-01-2021_03-16-24_PM

Maybe a rare situation but maybe might have to make use of the sibling `view_backward` in some cases.

Probably needs more sample data runs depending on the progress of the map. The case I encountered was 75% events were done however the newest one was further up yet it didn't auto locate to there, rather to the middle so close by but not visible.

nEEtdo0d:
I think PR is ready now. I've integrated the recommended changes and to some degree pruned `is_azur_affiliation` remaining place is the `ocr` for exchange count but at its core its using the color sampling like your other snippets which thank you very much for.

Stable as far as I can tell with running it but also done with missions for the week, the accept/collect will have to wait until tomorrow. Let me know if any further changes. Also if you feel like the lists of commits are too many I can start a new branch and migrate the changes from here to that one.

PR #321, opened 2021-01-26
Fix: Daily Ash and Guild Pop Up

nEEtdo0d:
- Daily Ash after combat may end up in unexpected `BEACON_MY`, allow recovery back into `BEACON_LIST`
- Guild pop up confirm match with regular pop up confirm implementation
- During campaign, `handle_in_map_with_enemy_searching` entered too early via `combat_status`, guild pop up may be encountered during this time and result in stuck status.  Add handling for this pop up during enemy search as well.
- During daily, guild pop up encountered too late and combat has already ended. May need to be re-written using `while` rather than `ui_click` and `for` loop


module/handler/enemy_searching.py
```diff
@@ -89,6 +89,12 @@ def handle_in_map_with_enemy_searching(self):
                 timeout.limit = 10
                 timeout.reset()
 
+            if self.handle_guild_popup_cancel():
+                self.config.GUILD_POPUP_TRIGGERED = True
+                timeout.limit = 10
+                timeout.reset()
+                continue
+
```
LmeSzinc, line 97: Code here doesn't effect daily, because the `expected_end` argument in daily combat is function `daily_end`, `handle_in_map_with_enemy_searching` is called when `daily_end` is None or equal `with_searching`.

To handle guild popup after daily, I suggest to add it to the argument `additional` in these two `ui_click()`
```
self.ui_click(click_button=button, check_button=self.combat_appear, appear_button=daily_enter_check,
                          additional=self.handle_combat_automation_confirm if not self.daily_auto_checked else None)
```
```
self.ui_click(click_button=BACK_ARROW, check_button=DAILY_CHECK)
```

nEEtdo0d: Sorry incorrect context for this change in code.  Related to this scenario that I encountered during a campaign run <https://pastebin.com/pC7yMMu5>, error folder contains this same image 



As for the daily encountering guild pop up, I should update both existing `ui_click`, the two you are suggesting are the ones marked by a inline comment (first/second alteration)?
```
        for n in range(remain):
            logger.hr(f'Count {n + 1}')
            self.ui_click(click_button=button, check_button=self.combat_appear, appear_button=daily_enter_check,
                          additional=self.handle_combat_automation_confirm if not self.daily_auto_checked else self.handle_guild_popup_cancel) # first alteration
            self.daily_auto_checked = True
            self.ui_ensure_index(fleet, letter=OCR_DAILY_FLEET_INDEX, prev_button=DAILY_FLEET_PREV,
                                 next_button=DAILY_FLEET_NEXT, fast=False, skip_first_screenshot=True)
            self.combat(emotion_reduce=False, save_get_items=False, expected_end=daily_end, balance_hp=False)

        self.ui_click(click_button=BACK_ARROW, check_button=DAILY_CHECK, additional=self.handle_guild_popup_cancel) # second alteration
```

`self.handle_guild_popup_cancel` should be accessible via `Daily` --> `Reward` --> `RewardX` --> `UI` --> `InfoHandler`

LmeSzinc: yes, update these two

nEEtdo0d:
Added `additional` parameter for daily, also noticed besides `daily`, `hard` can also encounter so ensured the `GUILD_POPUP_TRIGGERED` is retained through parent/calling object to later do `handle_guild`.  
- If SP/AB enabled because inherits `self.config` they will run and reset the config.
- If disabled then `reward_when_finished` will take care of it instead.

PR #332, opened 2021-02-06
Add: [EN] Add OS assets and fixes

nEEtdo0d:
- IN_MAP asset has been moved after update, fixes os_semi_auto
- handle_popup_single to handle pop up windows that only have a single button
- Add EN OS assets needed for os_clear
    - JP, TW copy common assets to reduce button_extract output. Only assets that are specific for server are printed.
- Update get_map_shape implementation using server indexing and attribute member for OCR object


nEEtdo0d:
06-02-2021_01-04-12_AM

Didn't notice among the added assets whether able to handle this sort of map.  Seems complicated so understandable, already glad is able to handle normal zones pretty well.  Thanks for your hard work!

nEEtdo0d:
Hadn't noticed but the IN_MAP asset changed location for this sort of zone. I wasn't really thinking but could probably have just used a different asset altogether but I went ahead and made a secondary asset IN_MAP_OBSCURED.

LmeSzinc:
I notice that there are map named `Méditerranée *` in EN, and I am wondering if OCR can detect the letter `é`.
There's no need to add another asset, you can use `offset` argument to handle assets moving

About other kinds of zones in operation siren, there won't be too many of them every month. So no need to auto them.

nEEtdo0d:
> I notice that there are map named `Méditerranée *` in EN, and I am wondering if OCR can detect the letter `é`.
I hadn't considered that.  But after using the module on couple of zones, the OCR check not super reliable I've had to move the camera several times before it finally got the right OCR value. Maybe consider having user enter the zone name themselves rather than doing OCR?

> There's no need to add another asset, you can use `offset` argument to handle assets moving
Gotcha but how would this be done in this case? Would it be 2 `self.appear` for one location and then another? I can remove the IN_MAP_OBSCURED references and swap Ash back to normal IN_MAP check as it did before.

> About other kinds of zones in operation siren, there won't be too many of them every month. So no need to auto them.
Semi-auto should still be supported right even for these zones right? But will need this `offset`

LmeSzinc:
> I hadn't considered that. But after using the module on couple of zones, the OCR check not super reliable I've had to move the camera several times before it finally got the right OCR value. Maybe consider having user enter the zone name themselves rather than doing OCR? Or maybe I just need to increase the threshold value...hadn't considered this. It was working so well with threshold=160 for so many sample images then live started not recognizing so well.

Reduce threshold to 96, because letters in EN is bolded. Typing map name is not a good idea.

> Gotcha but how would this be done in this case? Would it be 2 self.appear with one offset=x and the other having a different offset=y? I can remove the IN_MAP_OBSCURED references and swap Ash module back to normal IN_MAP check as it did before.
self.appear(IN_MAP, offset=x) or self.appear(IN_MAP, offset=y)?

`self.appear(IN_MAP, offset=(200, 5))` will handle both cases. It will search asset within range  (-200, -5, 200, 5).

> Semi-auto should still be supported right even for these zones right? But will need this offset handling

It can work ,but not recommended. If your fleet is just pass by the exit, semi-auto will click exit and confirm exit imediately, causing losses. Also, semi-auto will attack mob enemies if fleets accidentally encountered by an enemy in obscured zone. It's a waste of time to attack them.

nEEtdo0d:
> Reduce threshold to 96, because letters in EN is bolded.

Great, it does pull more consistently now but the whitespace is lost, so now `get_map_shape` must now be server specific which I re-arranged now. Thanks!

> Typing map name is not a good idea

Gotcha, I can see the pitfalls it might have had.

> `self.appear(IN_MAP, offset=(200, 5))` will handle both cases. It will search asset within range (-200, -5, 200, 5).

Works wonderfully, removed extra asset and adapted to using offset.  However I retained the ash.py change since `ui_click` uses the same `offset` for both `appear_button` and `check_button` so opted to still use a function.  Let me know if you want me to restore it completely and just have it do `ui_click(..., offset=(200, 5), ...)` instead

> It can work ,but not recommended. If your fleet is just pass by the exit, semi-auto will click exit and confirm exit imediately, causing losses. Also, semi-auto will attack mob enemies if fleets accidentally encountered by an enemy in obscured zone. It's a waste of time to attack them.

Ah right I had forgotten about that I just attacked them out of reflex.  Though with `offset=(200, 5)` they can use `os_semi_auto` should be careful in case an enemy is blocking their way perhaps or want to grab items along the way.

PR #340, opened 2021-02-11
fix: fix operation siren mediterranee map not found

kuwas:
Previous fix https://github.com/LmeSzinc/AzurLaneAutoScript/commit/1a3725a16c155d820629f92e4bc1a9db771d11de caused a mismatch between the get_map_shape parsed name and the DIC_OS_MAP defined name.


whoamikyo:
LGTM

PR #344, opened 2021-02-12
fix: fix campaign autosearch ignore low emotion stuck

kuwas:
When `IGNORE_LOW_EMOTION_WARN` is true and the campaign run is using autosearch, ALAS will ignore the popup as intended.

However, by ignoring this popup, Azur Lane will turn off autosearch in the background, even though combat has started correctly. This will cause ALAS to hang on the combat status screen since the code is actually expected AL to automatically skip this screen.

I've copied the status screen code from `combat.py` to address this in a quick and dirty way. Maybe someone who is more familiar with the code can clean it up and make everything more DRY though.

### DIFF
--- a/module/combat/auto_search_combat.py
+++ b/module/combat/auto_search_combat.py
@@ -168,6 +168,6 @@ class AutoSearchCombat(Combat):
 
         self.device.stuck_record_clear()
         self.auto_search_combat_execute(emotion_reduce=emotion_reduce, fleet_index=fleet_index)
-        self.auto_search_combat_status()
+        self.auto_search_combat_status(save_get_items=save_get_items)
 
         logger.info('Combat end.')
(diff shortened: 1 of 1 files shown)


LmeSzinc:
I think this is not the situation that auto-search is designed for. Auto search itself, just refuse to work. Alas need to click all the battle status.
Based on this, turn off auto search would be better. Not to mention that Alas is superior than auto search in choosing enemies and avoiding redundant mob battles before boss.

kuwas:
@LmeSzinc Except the current behavior looks like a bug, ALAS just gets stuck after combat and tries to restart the app over and over again. If this is intended behavior, at least make it configurable to the user how they want to proceed.

Also, the reason I'm using autosearch, is because it is significantly faster for zombie farming 2-1, which in turns leads to more gold and urgent commissions for oil and free gems. Based on what I've seen, even with the extra fights, autosearch is upwards of twice the speed of using the standard ALAS map search.

LmeSzinc:
I'm not responsible for extreme way to play. If you can implement it, and at the same time, not causing redudant clicks in normal situation, I will merge.

PR #358, opened 2021-02-17
Fix: guild_logistics and handle_mystery for os_clear_map

nEEtdo0d:
- guild_logistics too many clicks due to poor timing screen capture is actually pretty frequent for my pc so kept as personal change unless another reported something similar. In use for weeks personally, no issue as far as I can tell.
- Recent change in camera.py considers mystery encounter on path, os_clear_map can run into something similar with unexpected object on path as well. So optimized to handle those situations.

### DIFF
--- a/module/os/fleet.py
+++ b/module/os/fleet.py
@@ -66,7 +66,8 @@ class OSFleet(OSCamera, Combat, Fleet, OSAsh):
         return sight
 
     def handle_mystery(self, button=None):
-        return False
+        # Treat map events as mystery, may walk into unexpected object on path
+        return self.handle_map_event()
 
     def handle_ambush(self):
         # Treat map events as ambush, to trigger walk retrying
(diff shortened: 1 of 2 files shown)


module/os/fleet.py
```diff
@@ -66,7 +66,8 @@ def _walk_sight(self):
         return sight
 
     def handle_mystery(self, button=None):
-        return False
+        # Treat map events as mystery, may walk into unexpected object on path
+        return self.handle_map_event()
```
LmeSzinc, line 70: Map event is already handled in `handle_ambush()`, so here is redundant.
Also, if fleet encountered something in walk path, `handle_ambush()` will trigger a walk retry.
If it's handled in `handle_mystery()`, it won't trigger a walk retry, and finally cause walk timeout.

nEEtdo0d: Well I don't have the log anymore but I did indeed encounter an infinite loop of this printing statement that was added to camera.py.
```
elif self.appear(GET_ITEMS_1): logger.warning('Items got. Trying handling mystery.') self.handle_mystery() return self.update(camera=camera)
```
So it seemed like that was the best way to handle it. But if its as you say no problem. If I do encounter it again I will report and submit the error log next time.

PR #383, opened 2021-03-16
Upd: Reduce excessive error output for failed archives entrance search

nEEtdo0d:
- Add: [EN, JP, TW] Ink Stained Steel Sakura template asset
- Upd: Reduce excessive error output for failed search

Tried to replace with CN assets but encountered mixed results so instead cropped the Ink Stained Steel Sakura from CN to match format as other assets for EN, JP, and TW servers. Though the event is not available for EN, I hope this will be sufficient enough to support in the app update that unlocks it. May be re-visited if this template does not end up working.


LmeSzinc:
No need to copy the CN one.
Event banners are different among servers, although they have the same characters. For example the Nagato in EN banner might be smaller or bigger than the one in CN.
Confirm that most EN entrance can't work in CN, so I remaked all of them and cropped the whole faces for larger click area.

nEEtdo0d:
> Event banners are different among servers, although they have the same characters. For example the Nagato in EN banner might be smaller or bigger than the one in CN.
That explains why I was getting mixed results i thought the banners would be the same across servers just that the text area would differ. So pretty likely may need to revisit then in the future. Thanks for clarifying!

PR #390, opened 2021-03-24
Add: Option to use x2 Oil, XP, Drop plans

nEEtdo0d:
I wasn't confident in understanding how emotion.py works and how I could incorporate the plans detriment of double reduced emotion.  So worked from a different angle instead.

1. Enables the Efficiency Plan for selected campaign
2. Executes x number of campaigns
3. Rest for y minutes in reward loop
4. After y minutes, break out of loop
5. Repeat step 2

The main inclination why I wanted something like this is for war archives, seeing as data key plus plans is pretty good combo, still terrible drop rates but after 60 keys ~= 120 runs with the plans enabled I was able to make the D2 gun eventually. Also repeats the loop without having to start another session yourself.

One flaw I haven't been able to figure out is with the recent scroll and search for war archives campaign, I think sometimes the action to drag is read as a click, which exits the war archives interface, and ends up back in campaign menu so fails the search. This as a result disrupts the loop. Any way I could reduce the likelihood of this?

```            
p1, p2 = random_rectangle_vector(
    (0, -275), box=detection_area, random_range=(-50, -50, 50, 50), padding=20)
self.device.drag(p1, p2, segments=2, shake=(0, 25), point_random=(0, 0, 0, 0), shake_random=(0, -5, 0, 5))
```
I think possibly both `p1` and `p2` are randomly generated as really close in proximity to one another which ends up making the drag into a click instead. So maybe using some re-try on `random_rectangle_vector` to keep generating points until generating a `p1` and `p2` that are actually a distance a part? 

Another possibility may be the `self.device.drag` if I'm understanding it correctly it holds a position down and then moves but on the off chance can be read as a click instead perhaps due to emulator slow response or something. Would `self.device.swipe` be better to use then?


LmeSzinc:
Sorry to say that I can't accept this PR. The use of 2x book should be a setting in "setting", just like clear mode and auto search, instead of a standalone function.

Running and sleeping without precise calculation is a step backward. emotion.py should be refactored. I didn't imagine that emotion can be doubly reduced when I was writing it, so I hardcoded the reduction to 2. Reduction should depends on ENABLE_2X_BOOK (for example) and wait until it recover to the limit plus reduction.

Swipe distance of `random_rectangle_vector` depends on the argument `vector`, and it place the vector in `box` randomly. You can change the `box` to where it's save to click. I did manage to drop the closing swipes and swipe way points, but nothing changed.

The difference between `drag` and `swipe` is that, things still moves and then stop after a swipe, like having inertia. But drag stops quickly. When using uiautomator2, `drag` will shake at the end point (if moving vertically then shake horizontally). When using minitouch `drag` will stop at the end point for a while.

nEEtdo0d:
> Running and sleeping without precise calculation is a step backward. emotion.py should be refactored

Understood, yeah I got lost trying to dive into emotion.py.  But maybe I was overthinking it.  It seems you kind of already took into consideration, with the arguments `expected_reduce` = `2` or `(2, 2)`

So would just multiplying those or the constants used in emotion.py by 2 be sufficient? 

[emotion.zip](https://github.com/LmeSzinc/AzurLaneAutoScript/files/6200635/emotion.zip)
Not the most clever fix but it looks like just changing this would be enough. The default arguments though despite still using 2 are unaffected with how they are used in their calculations as far as I can tell. Like the `max(0, expected_reduce - self.const_reduce))` would still evaluate to 0, as its comparing 0 and -2. I'd be surprised if they made an item to reduce emotion any further than this. It also doesn't look like any of the functions being used across ALAS use anything but the default values besides `fleet`

Or would this screw up the bug_threshold check? Since its reducing at a faster rate, might even restart after 1 run. Then assumes emotion has recovered as a result and continues?

As an isolated test, changed restricted 5 loops to `while 1`. But surprisingly never exited war archives interface once. So perhaps not related `drag` . I guess it really just might be the my PC or emulator, registering some of ALAS actions as click maybe due to lag. Perhaps better resolution is to check if it accidently exited, recover by entering it again. May increase the try threshold in that case. The easiest might be to add `self.ui_ensure(destination=page_archives)` inside the loop, the print output would be repetitive but it would guarantee being in the interface at all times. Otherwise just encapsulating in `self.appear` base case check before calling it.

PR #393, opened 2021-03-27
Opt: Refactor emotion and add 2x Book config

nEEtdo0d:
Hopefully more aligned to your refactor idea for the emotion.py. Realized there were adjustments needed in other places and not just emotion.py. Do let me know if I overlooked something.

- Refactor emotion for 2x consumption due to item, only applicable if command is Main, Event, or War_archives
- Add 2x book config in level settings
- Support 2x book config enable in fleet preparation step, factor in odd cases
- Add 2x book assets in map's asset module
- Deleted redundant code for GUILD_POPUP_TRIGGERED, I'm the one who added this in the first place. Had no idea same object passed in is essentially a shared reference. Any updates to it then everyone that also has it can see it too. Little python familiarity. Made sense for `self.campaign.config` in doing this transfer since it is an independent copy that would not dynamically update.
- Bumped number of tries for war archives campaign search, recovery case if somehow exited page. I wouldn't be able to imagine the drag causing the exit 10 times in a row unless it really is a PC or emulator lag issue.


module/combat/emotion.py
```diff
@@ -143,5 +149,18 @@ def triggered_bug(self):
...
+        """
+        if self.config.ENABLE_2X_BOOK and \
+           self.config.COMMAND in ['Main', 'Event', 'War_archives']:
+            return 4
+        else:
+            return 2
+
+    def get_expected_reduce_reset(self):
+        """ Call this method if need to be reset. """
+        del self.__dict__['get_expected_reduce']
```
LmeSzinc, line 166: Use `@property` to get expected reduce in real time, not caching it.

module/map/map_operation.py
```diff
@@ -261,3 +262,32 @@ def handle_auto_search_emotion_wait(self):
...
+                backup.recover()
+            return False
+        elif self.config.ENABLE_2X_BOOK and self.appear(BOOK_ENABLE_CHECK) and self.emotion.get_expected_reduce == 2:
+            self.emotion.get_expected_reduce_reset()
+
+        enabled = self.image_color_count(BOOK_ENABLE_BOX, color=(156, 255, 82), threshold=221, count=20)
+        if (self.config.ENABLE_2X_BOOK and not enabled) or (not self.config.ENABLE_2X_BOOK and enabled):
+            self.device.click(BOOK_ENABLE_BOX)
+            self.device.sleep((0.5, 0.8))
+        self.ensure_no_info_bar()
```
LmeSzinc, line 293: Move this method to module/handler/fast_forward.py

No need to detect FLEET_PREPARATION, already checked before method called.

If enable 2x books, but map doesn't support 2x book (because haven't reached clear mode, or just don't have the option, etc), disable it.
Disable it won't effect the next run, config of each run is a fork of the original one.

Use `Switch` to ensure 2x book is enabled, a simple click is not enough.

nEEtdo0d: Sorry late reply

> Move this method to module/handler/fast_forward.py
> No need to detect FLEET_PREPARATION, already checked before method called.

Yes should be fine as this class inherits from there. The check was just used as a side precaution but really as long as no one uses it for any other screen interface, its fine.

> If enable 2x books, but map doesn't support 2x book (because haven't reached clear mode, or just don't have the option, etc), disable it.
> Disable it won't effect the next run, config of each run is a fork of the original one.

This I'm not too sure about is the config being referenced by the time it is using this `self.config` or `self.campaign.config`? If the prior, then its essentially a global update so changing it would be retained in all runs.  If the latter which is a fork as you say but its only forked once as far as I can tell via `self.load_campaign`, it didn't look like that config was reset at any point so each `self.campaign.run` call utilizes the same configurations as the last? 

So if changed once, wouldn't subsequent runs still retain `ENABLE_2X_BOOK = False`? The odd case I was hopefully trying to resolve is if 1st run couldn't turn it on however the 2nd run could, the emotion will be reduced 2 initially then 4 from then on. So using `cached_property` allowed it temporarily be 2 when combined with the `self.cover` and `self.recover` then back to 4 when the conditions were met.

Maybe I'm overlooking something? Thanks

> Use `Switch` to ensure 2x book is enabled, a simple click is not enough.

I can change to use this instead, I'll use the `def handle_auto_search` as a base

module/combat/emotion.py
nEEtdo0d: I can change to this but just wanted to first clear up whether I might be overlooking something. It doesn't look like each call to `self.campaign.run` uses a new forked config. Otherwise it would make sense to make this a `@property` since as you say `ENABLE_2X_BOOK = False` only for a single run then back to `True` after completion. If it still the map cannot support it, changing back to `False`. Repeat until doesn't need to keep resetting that config.

module/map/map_operation.py
LmeSzinc: Sorry that I forget that `load_campaign` only run once, but `self.campaign.config` is a copy of `self.config` and merge the config in map file. The `self.config` in fast_forward.py or map_operation.py is the `self.campaign.config` in campaign/run.py

2x books can be done like auto search. Add a flag `map_is_auto_search = False`, change the flag according to map info, not to change the original config `ENABLE_AUTO_SEARCH`.

nEEtdo0d: > 2x books can be done like auto search. Add a flag `map_is_auto_search = False`, change the flag according to map info, not to change the original config `ENABLE_AUTO_SEARCH`.

Okay I think I get it.  But I think this flag would need to be an attribute of `self.config` since emotion.py wouldn't be able to access it if defined in fast_forward.py. Something like the internal attribute `GUILD_POPUP_TRIGGERED`, call for now `EMOTION_2X`

Have `def handle_2x_book_setting` flip this either to `True` or `False` respectively based on the current conditions. emotion.py should be able to see this change via its `self.config` and use for `def get_expected_reduce`. Also can use for `Switch` to turn `on` or `off` if applicable. Hopefully I'm following your suggestion right.

Thanks

LmeSzinc: Add `self.map_is_2x_book` and `self.emotion.map_is_2x_book`, change them at the same time. I think the variables in self.config should be general configs and user configs which don't change frequently, and the variables in self are the infomation of map which change while running

nEEtdo0d: > Add `self.map_is_2x_book` and `self.emotion.map_is_2x_book`, change them at the same time. I think the variables in self.config should be general configs and user configs which don't change frequently, and the variables in self are the infomation of map which change while running

I think I gotcha, committed just now so I hope I got it right. Did some preliminary testing just making sure it checks and unchecks respectively depending on the config, also selected lethal maps to make sure whether book is actually available or not. Seems to work for EN at least. Also did logger prints to see `self.emotion.get_expected_reduce`

PR #401, opened 2021-04-09
Fix: Various After Update Bugs

nEEtdo0d:
- GET_MEOWFFICER_ITEM_* assets altered after update, posted new ones for EN. Other servers likely need to re-acquire as well
- Support page transition for Ink-Stained Steel Sakura War Archives for EN. TW and JP likely may need acquire their own as is the case for CN
- Extend handle_2x_book_setting for fleet prep and auto menu usage. Uses two different switch objects, pass appropriate argument as to which switch is used.  Alternatively, can move bulk to AutoSearchHandler instead if that is preferable since I wasn't too sure whether should remain in FastForwardHandler or not. Otherwise, `@Config ENABLE_2X_BOOK=True`?
- Fix transition to OS, immediately enters into map grid. Changed OS_CHECK to Ash Coordinate Logger icon. Page does not have concept of offset so don't think can use IN_MAP since position can vary based on current fleet's grid location.
- Add additional transition to daily assist via new 'Overview' button, use template to target location of click can vary based on current fleet's grid location


LmeSzinc:
Use `ui_click` in _ash_beacon_enter_from_map, set a large offset and it will handle the button moving

nEEtdo0d:
Apologies I did some re-arranging and converted using `while` loop page transition type rather than multiple `ui_click`

You are right, I had focused on thinking need template to handle button moving but a normal asset would do just fine with a large enough offset.

I think ready now, let me know if there is something else

Thanks

PR #411, opened 2021-04-21
Add: [EN] os and os_handler assets

nEEtdo0d:
In addition to counterpart assets, also includes:

- Simple bug fixes
- Update `handle_os_auto_search_map_option` to use new named asset  `AUTO_SEARCH_OS_MAP_OPTION_OFF` as EN unable to easily differentiate using same assets like CN
- Clone assets for JP and TW that are common or generic enough for all servers


nEEtdo0d:
With all pieces together, tried a complete run and worked well. So great job!

Just one small hiccup, happened at least 2 or 3 times so it took a few restarts. But specifically when shifting from St. Petersburg to Liverpool, I encountered this, from the screen shot trace, it seems like the `ZONE_ENTRANCE` click lags behind or times it just as it finally loads the maps. This may either click a grid to move to which seems to cause the port route to error or unintentionally switch the fleet if the button location happens to be close by. 
<https://www.dropbox.com/sh/dcmz1ymn132dpzf/AADHyi_D-NIgL4P6EgN-YCFwa?dl=0>

Perhaps `click_timer` could use the `count` argument seeing as its only given a duration or extend the duration in  `globe_operation.py` but 3 seconds is already quite long

As for the fleet switching, I think this might not be too problematic if the port dock repair becomes integrated. Is this supported? During the run, didn't look like it ran, got as low as 47% across main fleet HP so wasn't sure if there was a threshold or something. But otherwise, I think most people would only have 1 fleet with a healing CVL so accidental fleet switch could present to be a problem.

Happened only once so I doubt something to worry about. `ZONE_PINNED` resulted in too many click exception. I noticed the swipes are fairly small which is understandable, can't let it get too big. But I guess even after 10 tries some swipes are so small that the game still doesn't detect as zone not pinned.

Also question, does meowofficer farming have a stop condition or does it just loop forever until out of ap? I think adding a stop condition if ap window is encountered and don't want to use ap boxes or oil either, should stop too. Another one might be to farm until ashes daily is finished with greyed out 100/100.

nEEtdo0d:
> Increase the click interval when entering map to 10, because it may slow to load.

Your latest commit fixed this, increased the timer duration. Verified that I didn't run into transition issues moving from St. Peterburgs to Liverpool

> Sorry, I don't undertand what you mean. Can you send logs and screenshots to explain that

Sorry bad habit, I manually clean the logs and screenshots. But I do remember the code trace, I think root cause was here, it was `PINNED_DISABLE` not `ZONE_PINNED` apologies!  But again, only happened once I will upload if happens again.  But with latest commit, works stably, hasn't stopped for any exception during run. So really awesome! Basically, `self.is_zone_pinned()` kept evaluating to `True` so continuously swiped and eventually the too many click exception detected this.

```
if self.is_zone_pinned():
    # A click does not disable pinned zone, a swipe does.
    self.device.swipe((50, -50), box=area_pad(ZONE_PINNED.area, pad=-80), random_range=(-10, -10, 10, 10),
                          padding=0, name='PINNED_DISABLE')
```

> Loop until total AP (including boxes) lower than OS_ACTION_POINT_PRESERVE. 

I see I understand how this config works now. I was under the impression that it meant it would use a total of number preserve points and then stop after reaching that threshold. This way makes more sense this way less likely to have left over boxes at end of monthly reset.

> Adding a stop condiction to check if ash beacon is completed would be a good idea

Looks like pieces to try to support this have already been integrated. So perhaps something like this? Likely put something similar for obscure maps whenever that becomes supported. 

```
def os_meowfficer_farming(self, hazard_level=5):
    """
    Args:
        hazard_level (int): 1 to 6. Recommend 3 or 5 for higher meowfficer searching point per action points ratio.
    """
    logger.hr(f'OS meowfficer farming, hazard_level={hazard_level}', level=1)
    while 1:
        if self.config.ENABLE_STOP_FARM_ON_ASH_QUOTA and self._ash_fully_collected:
            break
        ...
```

PR #437, opened 2021-05-08
Fix: enhancement stuck

nEEtdo0d:
- Hadn't realized `wait_until_stable` uses `load_color` which can cause subsequent `appear` to fail for that asset even if working before. Now just using `wait_until_appear` instead

- Re-cut `CLICK_SAFE_AREA` to smaller hitbox for OS as `handle_map_get_items` is used to claim rewards, however original asset may click outside mission window and exit. Can perhaps be same asset across all servers if frequent enough, just from experience can happen occasionally


LmeSzinc:
I don't want click area to be too small, or change it because of one issue. The new `os_mission_enter` is able to re-enter mission list, when it's closed accidently.
A sleep before `wait_until_appear` is redundant, because wait_until_appear itself is already a wait

nEEtdo0d:
> I don't want click area to be too small, or change it because of one issue. The new `os_mission_enter` is able to re-enter mission list, when it's closed accidently.

Understood, this was encountered before the recent changes so perhaps you are right, if happens again I'll report. Thanks!

> A sleep before `wait_until_appear` is redundant, because wait_until_appear itself is already a wait

Agreed, removed sleep and left wait_until_appear. Thanks!

Also if you prefer I can make an entirely new branch with just the one change so that the commit history isn't bloated. Thanks!

LmeSzinc:
> Also if you prefer I can make an entirely new branch with just the one change so that the commit history isn't bloated. Thanks!

OK, make a new PR.

PR #445, opened 2021-05-15
Add: OS world clear module and hp/repair check support

nEEtdo0d:
Since the `os_world_clear` function already existed and having giving it go with some success, I thought it be nice to make it an actual module.  Had to make a few re-arrangements since `SAFE` maps could either be displayed immediately or in the selection list. Maybe not the cleverest way to go about it but made sense as I was working it.

Also with `fleet_repair` the only thing it really needed was a reason to be called. Initially, I had done zone clear count but then I figured HP checking made more sense. Got lost for a long while but eventually was able to find a happy medium, again maybe not the most clever way.

Thought I first post the PR before I migrate to other servers whether this could use more re-work or maybe just isn't your thing, which is fine. 

Honestly, as a temporary measure I'd probably do the same thing as `os_world_clear` but have it check for `OBSCURE` zones instead. Slower but at least will accomplish the intention until the real one is developed. I am a bit curious what you plan to do when you start working on it.


module/os/operation_siren.py
```diff
@@ -81,11 +82,13 @@ def globe_goto(self, zone, types=('SAFE', 'DANGEROUS'), refresh=False, stop_if_s
         self.globe_focus_to(zone)
         if stop_if_safe:
             pinned = self.pinned_to_name(self.get_zone_pinned())
-            if pinned == 'SAFE':
+            if pinned == 'SAFE' or \
+              (self.zone_has_switch() and self.zone_type_select(types='SAFE', exclusive=True)):
```
LmeSzinc, line 86: Add a new method `zone_has_safe` in globe_operation.py to replace this. Not to modify `zone_type_select`.

module/os/operation_siren.py
```diff
@@ -116,6 +119,17 @@ def fleet_repair(self, revert=True):
...
+        if enabled:
+            self.hp_get()
+            check = [round(data, 2) <= threshold if use else False for data, use in zip(self.hp, self.hp_has_ship)]
+            if any(check):
+                logger.info(f'At least one ship below configured threshold {threshold}, retreating to nearest azur port')
+                self.fleet_repair(revert=False)
+                self.hp_reset()
+            else:
+                logger.info('All ships fairly in good condition, continue OS exploration')
+
```
LmeSzinc, line 132: Add a new method `handle_fleet_repaire`, use ENABLE_OS_SETTING_REPAIR and OS_SETTING_REPAIR_THRESHOLD directly in it.

module/os/operation_siren.py
```diff
@@ -201,9 +217,11 @@ def os_meowfficer_farming(self, hazard_level=5, daily=False):
 
             self.globe_goto(zones[0])
             self.run_auto_search()
+            self.need_repair(self.config.ENABLE_OS_SETTING_REPAIR,
+                             self.config.OS_SETTING_REPAIR_THRESHOLD)
 
     def _clear_os_world(self):
-        for hazard_level in range(1, 7):
+        for hazard_level in range(self.config.OS_WORLD_MIN_LEVEL, self.config.OS_WORLD_MAX_LEVEL):
```
LmeSzinc, line 224: Add 1 to OS_WORLD_MAX_LEVEL here, not in reading config

LmeSzinc, line 224: Add 1 to OS_WORLD_MAX_LEVEL here, not in reading config

module/config/argparser_en.py
```diff
@@ -579,6 +579,26 @@ def choice_list(total):
...
+    os_world.add_argument('--os_world_max_level', default=default('--os_world_max_level'),
+                          help='No higher than 7\n'
+                               'Recommend 4 or lower for single fleet clear, 5 and higher '
+                               'can be difficult with low adaptibility',
+                          gooey_options={'label_color': '#4B5F83'})
+    os_world.add_argument('--enable_os_world_repair', default=default('--enable_os_world_repair'), choices=['yes', 'no'],
+                          help='Enable retreat to nearest azur port for repair if any one ship in fleet below configured threshold',
+                          gooey_options={'label_color': '#4B5F83'})
+    os_world.add_argument('--os_world_repair_threshold', default=default('--os_world_repair_threshold'), help='Limit between 0 and 1', gooey_options={'label_color': '#4B5F83'})
+
```
LmeSzinc, line 601: Need to emphasize that user need to clear the story mode in operation siren, buy and use 5k oil item.
Set `choices` argument in `os_world_min_level` and `os_world_max_level` to limit user input. I think any fleets over level 100 with Perseus and port repair is able to claer CL6.

LmeSzinc (commented):
> Honestly, as a temporary I'd probably do the same thing as os_world_clear but have it check for OBSCURE zones instead. Slower but at least will accomplish the intention until the real one is developed. I am a bit curious what you plan to do when you start working on it.

It will work like reward, record the last use time and trigger to clear a new one when it's free to use recon scan.

module/os/operation_siren.py
nEEtdo0d: So back out all changes in `zone_type_select`, including this part? 
```
        found = True # This found flag to determine whether entered the button is None clause
        for _ in range(3):
            self.ui_click(ZONE_SWITCH, appear_button=self.is_zone_pinned, check_button=self.is_in_zone_select,
                          skip_first_screenshot=True)

            selection = self.get_zone_select()
            logger.attr('Zone_selection', selection)
            button = get_button(selection)
            if button is None:
                logger.warning('No such zone type to select, fallback to default')
                types = ('SAFE', 'DANGEROUS')
                button = get_button(selection)
                found = False # Whether to return False since defaulted to a select
```
Since this flag is the important mechanism as I rely on the returning boolean whether actually found or just found something to take its place so can revert to pinned state.  Without it, based on testing always returned `True` since zone will likely always have the `DANGEROUS` option

Duplicate what I changed here and just place under `zone_has_safe` method signature?

module/os/operation_siren.py
nEEtdo0d: I don't use those configs directly in code since I had two sets (`ENABLE_OS_SETTING_REPAIR and OS_SETTING_REPAIR_THRESHOLD`) or (`ENABLE_OS_WORLD_REPAIR, OS_WORLD_REPAIR_THRESHOLD`) one for each module.  So based on which method was calling, it would have to change the configs.  Would you rather I reduce this to 1 set and only 1 module capable of configuring this so applies to all of them?

I thought it be more useful to have separate ones per module to have different settings but they can be combined if preferred

module/os/operation_siren.py
nEEtdo0d: No problem, easy adjustment

module/config/argparser_en.py
nEEtdo0d: Sure, I'll add in a requirements line before usage.

`choices` is easy adjustment and makes sense to limit the user.

Perhaps I should add when acquiring the zones of a certain hazard level to force fleet repair since for example if threshold isn't met after clearing an already relatively difficult zone then that would be bad as it be unlikely they could clear it.
```
for zone in zones:
    if not self.globe_goto(zone, stop_if_safe=True):
        continue
    self.run_auto_search()
    # self.need_repair(self.config.ENABLE_OS_WORLD_REPAIR,
    #                           self.config.OS_WORLD_REPAIR_THRESHOLD)
    handle_fleet_repair(force=True if hazard_level >= 5 else False)
```

nEEtdo0d:
While working on the hack for obscure zones I noticed the string for assets are mismatched `SELECT_OBSCURE` and `ZONE_OBSCURED`, was this intentional? Since I was thinking this might have been a typo as this would return 

`pinned = self.pinned_to_name(self.get_zone_pinned()) --> 'OBSURCED'` and to see whether obscure is selected you must pass in `self.zone_type_select(types='OBSCURE')`

nEEtdo0d:
Stuck state encounter before combat executed not sure how possible but somehow ended up in `AUTO_SEARCH_REWARD` screen before executing a combat

```
2021-05-16 16:31:48.686 | INFO | Click (1236,  581) @ AUTO_SEARCH_OS_MAP_OPTION_OFF
2021-05-16 16:31:57.606 | INFO | [Loading] 10%
2021-05-16 16:31:57.607 | INFO | [battle_count] 1
2021-05-16 16:31:57.607 | INFO | Auto search combat loading
2021-05-16 16:31:57.607 | INFO | Screenshot interval set to 1.0s
2021-05-16 16:32:14.636 | INFO | Auto Search combat execute
2021-05-16 16:33:18.770 | INFO | Screenshot interval set to 0.1s
2021-05-16 16:33:18.771 | INFO | Combat end.
2021-05-16 16:33:31.461 | INFO | [Loading] 10%
2021-05-16 16:33:31.461 | INFO | [battle_count] 2
2021-05-16 16:33:31.461 | INFO | Auto search combat loading
2021-05-16 16:33:31.461 | INFO | Screenshot interval set to 1.0s
2021-05-16 16:33:52.481 | INFO | Auto Search combat execute
2021-05-16 16:34:41.465 | INFO | Screenshot interval set to 0.1s
2021-05-16 16:34:41.466 | INFO | Combat end.
2021-05-16 16:34:50.577 | INFO | [battle_count] 3
2021-05-16 16:34:50.578 | INFO | Auto search combat loading
2021-05-16 16:34:50.578 | INFO | Screenshot interval set to 1.0s
2021-05-16 16:39:50.839 | WARNING | Wait too long
2021-05-16 16:39:50.840 | WARNING | Waiting for {'IN_MAP', 'SIREN_PREPARATION', 'BATTLE_PREPARATION', 'PAUSE'}
2021-05-16 16:39:50.840 | WARNING | Wait too long
```

Possible for some race condition to occur but now needs 2 break conditions. Ultimately, should still be able to exit since `is_in_map` but it technically didn't start any battle.
```
def auto_search_combat(self):
    ...
    while 1:
        self.device.screenshot()

        # Just in case maybe not necessary
        if self.handle_map_event():
            continue

        if self.is_combat_executing() or self.handle_os_auto_search_map_option():
            break
    ...
```

module/os/operation_siren.py
LmeSzinc: I think it's enough to use a same value.

module/os/operation_siren.py
LmeSzinc: It's not neccessary to call `zone_type_select`, You can implement like:
```python
def zone_has_safe(self):
    if self.pinned_to_name(self.get_zone_pinned()) == 'SAFE':
        return True
    else:
        self.ui_click(ZONE_SWITCH, appear_button=self.is_zone_pinned, check_button=self.is_in_zone_select,
                              skip_first_screenshot=True)
        flag = SELECT_SAFE in self.get_zone_select():
        self.ui_click(button, check_button=self.is_zone_pinned, offset=self._zone_select_offset,
                          skip_first_screenshot=True)
        return flag
```

module/config/argparser_en.py
LmeSzinc: I think it's not neccessary either. If you do want to check HP, check it after every combat, and stop auto search when lower than threshold.

LmeSzinc:
> I noticed the string for assets are mismatched `SELECT_OBSCURE` and `ZONE_OBSCURED`, was this intentional? The 'E' and 'ED' Since I was thinking this might have been a typo as this the following would be the case
> 
> `pinned = self.pinned_to_name(self.get_zone_pinned()) --> 'OBSCURED'` and to see whether obscure is selected you must pass in `self.zone_type_select(types='OBSCURE')` which affects its ability to break out of loop since `'OBSCURE' != 'OBSCURED'`
> 
> If is a typo, I can update the assets and comments wherever I find them so that they match

It's a typo

module/os/operation_siren.py
nEEtdo0d: After thinking more about it and maybe it is just a personal preference.

I'd like to consider keeping as it is if possible. I would think `os_world_clear` since its for clearing unsafe zones which has enemies in larger number. It is likely the threshold would be on the high end so that when its tackling the next zone then it is more likely to be prepared.  But for `os_fully_auto` I would imagine being on the lower end and used after `os_world_clear` in most cases I think it would be preferable doing dailies and farming when they are safe for a smoother run. Saving on changing the configuration too often based on the needs per module.

Also it saves the user from running a functional module and then stopping just to save a setting that affects another module behavior.

`need_repair` has been changed to `handle_fleet_repair', though the content hasn't really changed. But having it accept arguments on how it behaves I think is useful as it allows to use a fixed or configurable threshold. In this case, ensuring the fleet is ready however does not meet the configurable threshold before starting the module can be used as a precaution as exampled in the code.

However if you really don't agree, then I will make the adjustments accordingly for a single paired configuration. The question is which module you'd like the setting to be in `os_world_clear` or `os_fully_auto`? Or perhaps neither and instead added to the `settings` module if that is what you were suggesting

Sorry for the lengthy explanation and I hope it can considered but again I don't mind changing the infrastructure to single pair.

Thanks

module/config/argparser_en.py
nEEtdo0d: I think there may be 1 issue with checking after every battle especially when in auto search mode. Enemy nodes could be right next to each other and start a battle quite soon before ALAS could press the auto search button. 

It may easily get stuck with perhaps having enough time to detect hp but it already starts a new a battle. Making ALAS think, it can go fleet repair but the option to do that has disappeared now.

`hp_get()` screen capture may be taken during a loading screen or in battle thereby getting inaccurate reads as well.

While this would be easily remedied but I think should still be mentioned obscure zones, you wouldn't want to exit mid-run seeing as the zone disappears when you leave.

nEEtdo0d:
Besides the two additional comments and concerns I made, I think this PR may be ready to adapt for other servers like the `argparser` and `dictionary` py files. So if no additional changes is needed, I can move to that and then can be merged in full capacity.

Thanks

module/os/operation_siren.py
```diff
@@ -80,12 +81,12 @@ def globe_goto(self, zone, types=('SAFE', 'DANGEROUS'), refresh=False, stop_if_s
...
         if stop_if_safe:
-            pinned = self.pinned_to_name(self.get_zone_pinned())
-            if pinned == 'SAFE':
+            if self.zone_has_safe():
                 logger.info('Zone is safe, stopped')
                 self.ensure_no_zone_pinned()
                 return False
-        self.zone_type_select(types=types)
+        else:
+            self.zone_type_select(types=types)
```
LmeSzinc, line 89:
```python
else:
    self.zone_type_select(types=types)
```
Should revert to
```python
self.zone_type_select(types=types)
```
If stop_is_save is True, self.zone_has_safe() is False, and zone has dangerous and obscured zone, it will miss selecting zone and enter the default zone, which is obscured zone.

module/os/operation_siren.py
```diff
@@ -116,6 +117,21 @@ def fleet_repair(self, revert=True):
...
+                logger.info('At least one ship is below threshold '
+                           f'{str(int(threshold * 100))}%, retreating '
+                            'to nearest azur port for repairs')
+                self.fleet_repair(revert=revert)
+                self.hp_reset()
+            else:
+                logger.info('No ship found to be below threshold '
+                           f'{str(int(threshold * 100))}%, continue '
+                            'OS exploration')
+
```
LmeSzinc, line 134: Use the setting directly in this method.
```python
def handle_fleet_repair(self):
    if self.config.ENABLE_OS_SETTING_REPAIR:
        self.hp_get()
        check = [round(data, 2) <= self.config.OS_SETTING_REPAIR_THRESHOLD if use else False for data, use in zip(self.hp, self.hp_has_ship)]
        pass
```

module/os/operation_siren.py
```diff
@@ -221,6 +247,10 @@ def clear_os_world(self):
         # Force to use AP boxes
         backup = self.config.cover(OS_ACTION_POINT_PRESERVE=40, OS_ACTION_POINT_BOX_USE=True)
 
+        # Before start, repair if any ship is below
+        # a fixed threshold of 85% as precaution
+        self.handle_fleet_repair(True, 0.85, revert=False)
+
```
LmeSzinc, line 253: Call `self.handle_fleet_repair` and use the default settings would be enough.

module/os/operation_siren.py
LmeSzinc: I want to reduce the number of settings. Alas has much more setthings than other bots for game in 2D. There are too many settings for tiny things which are unneccessary.

When running `os_world_clear`, I assumed that users have perpared for it. They should have 1 or 2 healers in mob fleet. Alas is not responsable to satisfy people who want to attack zones in high CL with weak fleets, but Alas is able to continue running in that situation (Alas keeps turning on auto search and ships keep sacrifising themselves to kill enemies). So I think use the threshold in normal would be enough.

module/os/operation_siren.py
LmeSzinc: You can even delete ENABLE_OS_SETTING_REPAIR, fleet repair will be enabled when OS_SETTING_REPAIR_THRESHOLD > 0
```python
def handle_fleet_repair(self):
    if self.config.OS_SETTING_REPAIR_THRESHOLD > 0:
        self.hp_get()
        check = [round(data, 2) <= self.config.OS_SETTING_REPAIR_THRESHOLD if use else False for data, use in zip(self.hp, self.hp_has_ship)]
        pass
```

module/os/operation_siren.py
nEEtdo0d: I think this case is already handled? Or I might be missing something? `zone_has_safe` will choose `SELECT_SAFE` if present also returning `True` otherwise `SELECT_DANGEROUS` since that is available in every zone and returns `False`? But it shouldn't need to call `zone_type_select` since it already made the necessary selection. Unless you want it to go through the selection process again? But I would think not?

```
def zone_has_safe(self):
    if self.pinned_to_name(self.get_zone_pinned()) == 'SAFE':
        return True
    else:
        self.ui_click(ZONE_SWITCH, appear_button=self.is_zone_pinned, check_button=self.is_in_zone_select,
                            skip_first_screenshot=True)
        flag = SELECT_SAFE in self.get_zone_select()
        button = SELECT_SAFE if flag else SELECT_DANGEROUS
        self.ui_click(button, check_button=self.is_zone_pinned, offset=self._zone_select_offset,
                        skip_first_screenshot=True)
        return flag
```

The alternative would be to make `zone_has_safe` to making no selection and changing `zone_type_select` to handle when in the zone menu/option or in the pinned state where the zone switch button can be clicked but since it uses `ui_click` primarily that may be difficult.

module/os/operation_siren.py
nEEtdo0d: Understood, fair point. I shouldn't have problem reducing this is just 1 setting and will be placed in `os_fully_auto` along with the other settings.

Thanks

module/os/operation_siren.py
nEEtdo0d: This is true, just using threshold should be sufficient and play the role as both enabler and configurable.

Thanks

module/os/operation_siren.py
nEEtdo0d: Understood, it will just use configured settings. No need for precaution.

nEEtdo0d:
With this last commit I hope I've addressed the last. If everything is well, I'll make the last commit with argparser/dictionary updates for servers to introduce the new settings and module.

module/os/operation_siren.py
nEEtdo0d: I did realize one hindsight though, in the case there is no `zone_switch` so I've updated to
```
def zone_has_safe(self):
    if self.pinned_to_name(self.get_zone_pinned()) == 'SAFE':
        return True
    elif self.zone_has_switch():
        pass
    else:
        # No zone_switch, already on DANGEROUS
        return False
```

Should be the case if no switch is present at all then its `SELECT_DANGEROUS`, no?
Otherwise in the most minimal case (`SELECT_DANGEROUS` and `SELECT_OBSCURE` the 2nd clause will appropriately `SELECT_DANGEROUS`

LmeSzinc: I think `zone_has_safe` is a method to get infomation from game, and `zone_type_select` is a method to do operations in game. To prevent making thing complecated. A method to get infomation shouldn't do operations in game, vice versa.
`zone_type_select` itself can skip selecting if it's already at targer genre.

nEEtdo0d: Understood, the nomenclature should be followed, Thought it save a few clicks and less hassle in doing things twice. I've removed the  `else` and will move onto the server files next. Thanks

nEEtdo0d:
Last commit hopefully (haha), long run out of non-SAFE zones so substituted using other zone types to test the behavior. I think pretty stable.

Thanks for sticking with me on this pull request.

PR #487, opened 2021-06-14
Fix: 2x book enable in fleet prep mode, likewise done in auto mode

nEEtdo0d:
Fleet prep mode for some reason affected same way, it didn't before not sure why.  So now both 2x book options in either interface works consistently regardless of what type map is being entered.


LmeSzinc:
I think you need to deside whether to use `offset` or not. Not to change it back and forth.

nEEtdo0d:
Yeah sorry for the inconsistencies, I had been using without issue before the event then suddenly getting reports of not working.

Pulled at straws for a while but it wasn't until I removed `offset` that it fixed. It was strange since I thought `offset` worked as a means to be lenient so long as the asset appears within a closed range so doesn't feel like the `self.appear` in `Switch` should have returned `False` with `offset` but it did.

PR #496, opened 2021-06-24
Fix: 2x book verify and apply check mechanism with limited retry

nEEtdo0d:


LmeSzinc:
I think some implovements can be made.
1. 4 assets (BOOK_CHECK_PREP, BOOK_BOX_PREP, BOOK_CHECK_AUTO, BOOK_BOX_AUTO) can be merged in to 2 assets. 
    You can use large offset and call `self.appear()`, then the `button` attribute is the real corrdinates of the asset. So 1 asset is able to detect both `BOOK_CHECK_PREP` and  `BOOK_CHECK_AUTO`.
    `BOOK_BOX_*` is related to `BOOK_CHECK_*`. Use `BOOK_BOX.load_offset(BOOK_CHECK)`, and then the `button` attribute of `BOOK_BOX` is the real coordinates too.
2. Increase click interval to 3 seconds, because the click animation may mess up color detection. Retry is rarely called, so large interval won't slow things down.
3. Delete `_ensure_2x_book_appear`. You can use `if self.appear(BOOK_CHECK, offset) and clicked_timeout.reached():` and it will click it when 2x book setting is appear.

nEEtdo0d:
Some of these I'm not too sure whether they might work as intended
1. It would be nice to merge the assets however I think there might be an issue for `BOOK_CHECK_*`. Comparing the two book icons in the prep menu vs the auto menu, the resolution/size of the icons differ and may result in not quite detecting it.
https://imgur.com/a/2FU3RcA
If we use the text instead of the book, the font size also differ maybe not as much as the book icon though. But I chose the book since it would be fairly transparent across servers assuming it hadn't moved in position (unfortunately some assets do like some of the auto search assets).
The `BOOK_BOX_*` asset I'm not seeing how to merge into one seeing as their check boxes are at two different positions.
Is `load_offset` used to emulate a sort of `.BUTTON` effect where `self.appear(BOOK_BOX_*)` would really be doing `self.appear(BOOK_CHECK_*)` but `self.device.click(BOOK_BOX_*)` would still click wherever `BOOK_BOX_*` is positioned at?
2.  and I think 3. go hand in hand and I think both these are do-able. A minor alteration to your suggestion but I hope this next check in is sort of what you had in mind.

Thanks for reviewing!

module/handler/fast_forward.py
```diff
@@ -190,27 +185,86 @@ def triggered_map_stop(self):
...
+                skip_first_screenshot = False
+            else:
+                self.device.screenshot()
+
+            if clicked_threshold < 0:
+                break
+
+            if clicked_timeout.reached():
+                if self.appear(check_button):
+                    enabled = self.image_color_count(box_button, color=(156, 255, 82), threshold=221, count=20)
```
LmeSzinc, line 226: Set a large offset in `self.appear(check_button)`, such as (100, 50), so `BOOK_CHECK_*` don't need to be very among servers.

module/handler/fast_forward.py
```diff
@@ -190,27 +185,86 @@ def triggered_map_stop(self):
...
+                    if ((status == 'on' and enabled) or (status == 'off' and not enabled)):
+                        return True
+                    if ((status == 'on' and not enabled) or (status == 'off' and enabled)):
+                        self.device.click(box_button)
+
+                clicked_timeout.reset()
+                clicked_threshold -= 1
+
+        logger.warn(f'Wait time has expired; Cannot set 2x book setting')
+        return False
```
LmeSzinc, line 236: Use `logger.warning()` instread of `logger.warn()`.
https://stackoverflow.com/questions/15539937/whats-the-difference-between-logging-warn-and-logging-warning-in-python/15655674

module/handler/fast_forward.py
```diff
@@ -190,27 +185,86 @@ def triggered_map_stop(self):
...
+            if clicked_threshold < 0:
+                break
+
+            if clicked_timeout.reached():
+                if self.appear(check_button):
+                    enabled = self.image_color_count(box_button, color=(156, 255, 82), threshold=221, count=20)
+                    if ((status == 'on' and enabled) or (status == 'off' and not enabled)):
+                        return True
+                    if ((status == 'on' and not enabled) or (status == 'off' and enabled)):
+                        self.device.click(box_button)
```
LmeSzinc, line 230: Reduce one layer of parentheses。
`if (some_complex_conditions) or (some_complex_conditions):` is enough.

module/handler/fast_forward.py
```diff
@@ -190,27 +185,86 @@ def triggered_map_stop(self):
...
             self.map_is_2x_book = False
             self.emotion.map_is_2x_book = self.map_is_2x_book
-            return False
-
-        self.emotion.map_is_2x_book = self.map_is_2x_book
-        status = 'on' if self.map_is_2x_book else 'off'
 
-        book.set(status=status, main=self)
         self.ensure_no_info_bar()
+        return True
```
LmeSzinc, line 270: Use `self.handle_info_bar()` instead of `self.ensure_no_info_bar()`.
Info bar only appears when option is changed. If option is correct, handle_info_bar can save some time.

LmeSzinc (commented):
All good, just some minior code styles.

nEEtdo0d:
Good point, we can still use `offset` even still with two assets to account for server differences for positioning.

Applied minor code styles as commented, thanks!

Stable for EN and hopefully the last time this get re-visited.

PR #508, opened 2021-07-02
Add: Reward Shop module

nEEtdo0d:
This was challenging, ran into a lot of trial by fire testing but works with low error rate I think.

Handling purchases in the second grid for shop guild took quite a bit of planning. Didn't go with using `ItemGrid` as they are fixed locations and some items using the same template, unable to differentiate them.

The shop medal was the most difficult as I often ran into failing match templates and required fiddling with the `ButtonGrid` several times. I'm not positive but that grid in particular seemed inconsistent day by day with the pixels not always at the `origin` or just being slightly off the next `delta` item.  I tried a range of different parameters before settling with the current even now might not capture all items, low chance but can happen.

I saw in your gems_farming branch, a few upgrades to some modules, so will update those affected areas if that goes in first.

commit: 431911d51f5cc439b581beceb216bdf7adab87b6 lists all the assets I think other servers would need to acquire their own copy of.

EDIT:
I hadn't even considered about these constants in `ItemGrid`, rather then just fiddling with `ButtonGrid`...
Would lowering this to maybe like `0.85` be okay to change or too low. Some items look too similar so might classify some items wrong. The closest I've gotten has been `0.90` on those failed templates.
```
class ItemGrid:
    similarity = 0.92
    ....
```


LmeSzinc:
This is a lot of work, thanks for the contribution.

Add a new class to wrap *_bar_click and *_bar_ensure. Take Switch and Scroll as examples. Those code are frequetly used, needs to become a general handler.
I changed the dev brunch, will add CN assets and test later.

nEEtdo0d:
> This is a lot of work, thanks for the contribution.

If it looks within my capability, I have fun with it and glad I can contribute. This one started strong however definitely got painful towards the end haha.

> Add a new class to wrap *_bar_click and *_bar_ensure. Take Switch and Scroll as examples. Those code are frequetly used, needs to become a general handler.

There are parts that can be generalized but some areas had to be re-adjusted so I'll see if I can prototype something, no guarantees on success.

nEEtdo0d:
Sorry took a break for holidays but I think made a pretty successful implementation using a common class to do the side/bottom bar interactions.

Just the shop and build modules adopted this style, if you think this looks satisfactory I can adopt the same style for the other modules equipment and guild I think are the other ones in this case.

LmeSzinc:
PageBar didn't simplify the usage, try this:
<details>
  <summary>click to expand</summary>

```python
from module.base.button import ButtonGrid
from module.base.base import ModuleBase
from module.logger import logger
from module.base.timer import Timer


class Navbar:
    def __init__(self, grids, active_color=(247, 251, 181), inactive_color=(140, 162, 181), active_threshold=180,
                 inactive_threshold=180, active_count=100, inactive_count=50, name=None):
        """
        Args:
            grids (ButtonGrid):
            active_color (tuple[int, int, int]):
            inactive_color (tuple[int, int, int]):
            active_threshold (int):
            inactive_threshold (int):
            active_count (int):
            inactive_count (int):
            name (str):
        """
        self.grids = grids
        self.active_color = active_color
        self.inactive_color = inactive_color
        self.active_threshold = active_threshold
        self.inactive_threshold = inactive_threshold
        self.active_count = active_count
        self.inactive_count = inactive_count
        self.name = name if name is not None else grids._name

    def get_info(self, main):
        """
        Args:
            main (ModuleBase):

        Returns:
            int, int, int: Index of the active nav item, leftmost nav item, and rightmost nav item.
        """
        total = []
        active = []
        for index, button in enumerate(self.grids.buttons):
            if main.image_color_count(
                    button, color=self.active_color, threshold=self.active_threshold, count=self.active_count):
                total.append(index)
                active.append(index)
            elif main.image_color_count(
                    button, color=self.inactive_color, threshold=self.inactive_threshold, count=self.inactive_count):
                total.append(index)

        if len(active) == 0:
            logger.warning(f'No active nav item found in {self.name}')
            active = None
        elif len(active) == 1:
            active = active[0]
        else:
            logger.warning(f'Too many active nav items found in {self.name}, items: {active}')
            active = active[0]

        if len(total) < 2:
            logger.warning(f'Too few nav items found in {self.name}, items: {total}')
        if len(total) == 0:
            left, right = None, None
        else:
            left, right = min(total), max(total)

        return active, left, right

    def get_active(self, main):
        """
        Args:
            main (ModuleBase):

        Returns:
            int: Index of the active nav item.
        """
        return self.get_info(main=main)[0]

    def get_total(self, main):
        """
        Args:
            main (ModuleBase):

        Returns:
            int: Numbers of nav items that appears
        """
        _, left, right = self.get_info(main=main)
        return right - left + 1

    def set(self, main, left=None, right=None, upper=None, bottom=None, skip_first_screenshot=True):
        """
        Set nav bar from 1 direction.

        Args:
            main (ModuleBase):
            left (int): Index of nav item counted from left. Start from 1.
            right (int): Index of nav item counted from right. Start from 1.
            upper (int): Index of nav item counted from upper. Start from 1.
            bottom (int): Index of nav item counted from bottom. Start from 1.
            skip_first_screenshot (bool):

        Returns:
            bool: If success
        """
        if left is None and right is None and upper is None and bottom is None:
            logger.warning('Invalid index to set, must set an index from 1 direction')
        text = ''
        if left is None and upper is not None:
            left = upper
        if right is None and bottom is not None:
            right = bottom
        for k in ['left', 'right', 'upper', 'bottom']:
            if locals().get(k, None) is not None:
                text += f'{k}={locals().get(k, None)}'
        logger.info(f'{self.name} set to {text}')

        interval = Timer(2, count=4)
        while 1:
            if skip_first_screenshot:
                skip_first_screenshot = False
            else:
                main.device.screenshot()

            active, minimum, maximum = self.get_info(main=main)
            if active is None:
                continue
            index = minimum + left - 1 if left is not None else maximum - right + 1
            if not minimum <= index <= maximum:
                logger.warning(
                    f'Index to set ({index}) is not within the nav items that appears ({minimum}, {maximum})')
                continue

            # End
            if active == index:
                return True

            if interval.reached():
                main.device.click(self.grids.buttons[index])
                main.device.sleep((0.1, 0.2))
                interval.reset()
```
</details>

Usage: Build event pool if available, if not available build light pool
```python
self = ModuleBase('alas')
self.device.screenshot()
BUILD_SIDEBAR.set(main=self, upper=1)
if BUILD_BOTTOMBAR.get_total(main=self) == 4:
    BUILD_BOTTOMBAR.set(main=self, left=1)
else:
    BUILD_BOTTOMBAR.set(main=self, right=3)
```

nEEtdo0d:
Whoa impressive writing this up so quickly and thanks! I'll do some trial runs before migrating it completely but seems to mesh pretty well from a glance. Been swamped lately so may not get to it super soon. 

Though there are some parts of your original implementation which were handy. Like keeping indexes the same for certain navigator clicks even if the button grid was found smaller or larger than expected.  Build pool is a good example, regardless whether the event pool was there or not, light pool would always be 2. Though I can understand the requirement for the `additional` parameter that PageBar had made it more complex.

Minor but saw this requires at least one of the buttons to be active within view. For a grid like shop, you can end up with a view with none being active. Not a big deal, just means may have to use the alternative transition path to munitions. Going via the page academy guarantees arriving in the general shop and using the shop button path defaults to the right most shop whatever that might be. The backgrounds of those limited shops can significantly change the active or inactive colors, so taking the longer path may be better in the long run.

nEEtdo0d:
Sorry for the delay, Navbar added! Adopted Navbar to all relatable modules now.

Ran independent scenarios so running them all together in bulk I haven't tried yet but pretty confident, I still make mistakes though.

The commit history is a mess I know so apologies for that.

PR #515, opened 2021-07-08
Add: [EN] raid, meowfficer, 2x book, and auto search assets

nEEtdo0d:
* Add: [EN] raid ocr and entrance assets

First used daily module to run it rather than the standalone raid module. It used Essex ocr asset strangely. Looking into it, saw that it uses RAID_NAME rather than RAID_DAILY_NAME. So made sense since I hadn't run the standalone, my setting was still set to Essex at the time. But after manually setting that, the daily module worked after that.

Maybe the daily settings for raid can be converted to how os_fully_auto enables for daily instead since it looks like RAID_DAILY_NAME is now deprecated. Also the standalone is set up to run 1 difficulty at a time.

* Add: [EN] meowfficer comf-forts info handling for page_meowfficer


nEEtdo0d:
Not sure how to handle this scenario, recent update had moved `MEOWFFICER_TRAIN_START` a few pixels from its original location so now, this part gets stuck (line 250, reward/meowfficer.py)
`self.ui_click(MEOWFFICER_GOTO_DORM,
                      check_button=MEOWFFICER_TRAIN_ENTER, appear_button=MEOWFFICER_TRAIN_START, offset=None)`

So easy fix, remove the `offset` or so I thought,  But now we get a different problem. The `ui_click` exits prematurely since default `offset=(20, 20)` is used across both assets. Despite `MEOWFFICER_TRAIN_ENTER` being greyed out this evaluates to `True` so it didn't actually do anything. Don't mind the extra logger prints.
```
2021-07-08 18:12:16.761 | INFO | Click (1037,  566) @ MEOWFFICER_TRAIN_START
2021-07-08 18:12:21.501 | INFO | about to ui_click
2021-07-08 18:12:21.502 | INFO | False
2021-07-08 18:12:21.502 | INFO | True
2021-07-08 18:12:21.503 | INFO | True
2021-07-08 18:12:21.503 | INFO | <<< UI CLICK >>>
2021-07-08 18:12:21.729 | INFO | exit prematurely due to default offset
2021-07-08 18:12:21.729 | INFO | <<< UI ENSURE >>>
2021-07-08 18:12:21.956 | INFO | [UI] page_meowfficer
2021-07-08 18:12:21.956 | INFO | Goto page_main
2021-07-08 18:12:21.957 | INFO | [UI route] page_meowfficer - page_main
2021-07-08 18:12:21.957 | INFO | <<< UI CLICK >>>
2021-07-08 18:12:22.162 | INFO | Click (  42,   31) @ MEOWFFICER_GOTO_DORM
```

The good thing is that `ui_goto_main` follows up and is able to handle this.  Although now it takes 10 seconds with 2 clicks to exit altogether because first click exits the training window and the second does the actual page transition. It also uses the default offset so makes sense as to why it didn't have trouble seeing `MEOEFFICER_TRAIN_ENTER`

How to proceed from here:
- Should `MEOWFFICER_TRAIN_START` just be re-cut? Maintaining the same line as it is now.
- Should `ui_click` have more customization with separate offset configurable for the two assets? 
- Don't use `ui_click` at all, maybe new shared method explicitly for exiting back window of either buy or train

Thoughts?

PR #524, opened 2021-07-10
Add: confirm_wait optional parameter for ui_ensure/ui_goto (tentative)

nEEtdo0d:
The main topic is `page_meowfficer` or `page_dorm` still often lags behind on `*_INFO` pop ups. The `*_CHECK` assets seem to load first and with having `confirm_wait=1` by default I think is where the problem might be since those `*_INFO` tend to appear `> 1`.

Sure, it adds more delay now but now it has better chance encountering those `ui_additional_*` funcs. I had wondered why `_dorm_receive` had special circumstances when really those funcs should be handlng it.

The only scenario that uses `confirm_wait > 1` is for withdrawing from a campaign, so since there is case to use it. Extending the usage to these pages I think is reasonable.

But if you think this just impedes ALAS too much, I understand.

### DIFF
--- a/module/reward/dorm.py
+++ b/module/reward/dorm.py
@@ -227,7 +227,7 @@ class RewardDorm(UI):
             if not feed:
                 self.ui_goto_main()
                 return
-        self.ui_goto(page_dorm, skip_first_screenshot=True)
+        self.ui_goto(page_dorm, confirm_wait=3, skip_first_screenshot=True)
 
         if collect:
             self._dorm_receive()
--- a/module/reward/meowfficer.py
+++ b/module/reward/meowfficer.py
@@ -264,7 +264,7 @@ class RewardMeowfficer(UI):
         if not buy and not train:
             return False
 
-        self.ui_ensure(page_meowfficer)
+        self.ui_ensure(page_meowfficer, confirm_wait=3)
 
         if buy:
             self.meow_buy()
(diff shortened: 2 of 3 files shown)


LmeSzinc:
I think 3 seconds is too long. Perhaps we should find another way to solve it, instead of using confirm timer.

nEEtdo0d:
> I think 3 seconds is too long. Perhaps we should find another way to solve it, instead of using confirm timer.

Initially was just thinking of using the typical wait/confirm loop used everywhere but when I saw confirm_timer in ui_click, it seemed like good idea as it already does exactly that plus is part of the page transition mechanism already.

Implementing like 2x book check to handle both fast/slow cases but I feel like that would also gloss over the `*_INFO` if its particularly slow to load as the `*_CHECK` assets show up first.

I don't think in either of those pages have an asset that can be used as a guarantee that only loads after `*_INFO`

Sample logs with the alterations from this PR as this gives a vague gauge the duration my emulators took to see those `*_INFO` windows.
https://pastebin.com/s9pTiR1Q

nEEtdo0d:
After a few trial runs, took a different approach and this seems to allow sufficient amount of delay without hindering other servers and altering `confirm_wait` for everyone

https://github.com/LmeSzinc/AzurLaneAutoScript/pull/539

PR #555, opened 2021-08-01
Add: shop module missing components and shipyard module support for daily discounted research prints

nEEtdo0d:
Even when using the nifty diff tool I forgot a few things in that previous PR that was my bad, sorry. But functionally ALAS was fine without them being there as I was using these changes regularly.

So PR contains three parts:
- merges latest from master
- shop module missing build module assets across servers, dictionary.py updates for shop related terminology
- shipyard module; configuration interface, config.py, dictionary.py, assets, reward add on and etc.

Could have made two PRs for the last 2 but was testing all as one cohesive change and seems to be successful at least for EN

dev only edits like `buttons()` to `buttons` already in this PR.  Still using `buttons()` for my personal branch though.

It's a big PR so I get it, might take a few weekends to sort things out. I'll be patient this time around.


LmeSzinc:
Your code is trustworthy now, so I will pay attention on overall directions instead of implementation details. I send you an inviation to be a collaorator of ALAS. By accepting it, you can push to the this repo directly. Please check your email, and deside whether to accept or reject.

1. I would like to rename `module/build` to `module/gacha`, because `/build` is in `.gitignore`. Sure, you can delete that in gitignore, that's another way to solve it.
2. `_shipyard_ensure_index` can be replaced with `ui_ensure_index`, it seems to be redundant.
3. `_shipyard_set_index` can be replaced with `ui_click`

nEEtdo0d:
> Your code is trustworthy now, so I will pay attention on overall directions instead of implementation details. I send you an inviation to be a collaorator of ALAS. By accepting it, you can push to the this repo directly. Please check your email, and deside whether to accept or reject.

Oh wow thanks! I will accept but of course I'll try to be as thorough as possible with what I'm working on before finalizing it.

> 1. I would like to rename `module/build` to `module/gacha`, because `/build` is in `.gitignore`. Sure, you can delete that in gitignore, that's another way to solve it.

I gotcha, didn't notice that. This might be how it was missed  since I tend to use the TortoiseGit GUI to make commits. No problem, I'll fit it into this PR. Just the module name is enough? Or also the assets like `BuildUI` to `GachaUI` and `build_navbar` to `gacha_navbar`?

> 2. `_shipyard_ensure_index` can be replaced with `ui_ensure_index`, it seems to be redundant.

I tried using `ui_ensure_index` before but because it doesn't exit if it cannot press the number of times given, it gets trapped. I was getting overwhelmed how to calculate the number of prints needed whether it was DEV or FATE since the OCR for both were different; location wise, counters vs percentages, and needing less than configured to reach max level. The total number of prints required for every level is posted in the wiki but transcribing that information to the module didn't seem productive.

So went with a different route with how many can be used by clicking first then OCR. Put that number against the configured to calculate a remainder, loop again if `> 0`. This way it can handle transitions to DEV and FATE i.e. need 1 print before DEV finishes and then use the remaining free 1 print towards the FATE contribution. If configured with `> 2` then uses the remainder all towards FATE except for the 1 that was needed by DEV.

> 3. `_shipyard_set_index` can be replaced with `ui_click`

I didn't use `ui_click` because I didn't find any assets good for `check_button` and `appear_button`.  The button being clicked is generated by `ButtonGrid` so it being `appear_button` wouldn't have worked either.  Also everything remained on the screen whether or not the face button was clicked, `ui_click` can prematurely exit as a result.  Only when a new face button was clicked, would the assets recede and come back to reflect the new selection.

These are the raws I used to cut out the EN assets
https://imgur.com/a/yv5yz5d

LmeSzinc:
1. Seems ready to go
2. OK, keep `_shipyard_ensure_index`.
3. You can use `offset=None` which will use color match and only needs `area`  and `color`
Or wrap detection in a function, i.g. module/daily/daily.py Daily.daily_execute()
```
def daily_enter_check():
    return self.appear(DAILY_ENTER_CHECK)
self.ui_click(click_button=DAILY_ENTER, check_button=daily_enter_check, appear_button=DAILY_CHECK,
                   skip_first_screenshot=True)
```

nEEtdo0d:
https://pastebin.com/2vWPNTKW tried using `offset=None` with the generated button but that resulted in an exception.

Definitely butchered how `ui_click` was meant to be used but I wasn't having much luck with either suggestion but these technically work. Less code for sure but pretty confusing unless stepping through.
https://pastebin.com/Mc778R7Z

It's actually `check_button` and `confirm_wait` that makes `ui_click` exit early. I could set it up like this technically too but this doesn't seem like a productive set up either.
```
self.device.click(button)
self.ui_click(...)
```

As I said, the interface being handled doesn't really change in appearance or go to a different page. Besides the ship girl in the center, all other assets in this are all pretty much static, only disappearing momentarily in case a new ship girl is selected instead.

Made a gif giving the idea, I don't think there is suitable `check_button` or `appear_button` that can be used for this handling.  Also can't depend on the ship girl itself like how `enhancement` uses for checking a different ship, in the case the default selected is the one the user wants to buy the BPs for.

LmeSzinc:
I think there should be a `Navbar` object to ensure it is selected to the correct index. You can detect the background in blue/darkblue.
SHIPYARD_UI_CHECK appears before and after click, so if click was dropped, current `_shipyard_set_index` function may think it's correct and exit. `ui_click` cannot make a click either, because `appear_button` and `check_button` are the same.

nEEtdo0d:
> I think there should be a `Navbar` object to ensure it is selected to the correct index. You can detect the background in blue/darkblue.

I had considered `Navbar` when I had first started as well but the faces in the grid covered different parts within its grid space, so I was afraid there might not be one spot in that grid that I could 100% depend on always free or not blocked being where I can sample the active or inactive color consistently.  I know we only have PR1 currently, but I presume other series will be available in the future so then this would need to revisited in case a different series didn't like that as a good common spot for the `Navbar`.

If I take a really small corner area, I can try to draft something up.

> SHIPYARD_UI_CHECK appears before and after click, so if click was dropped, current `_shipyard_set_index` function may think it's correct and exit. `ui_click` cannot make a click either, because `appear_button` and `check_button` are the same.

This is true, I hadn't considered that. It's really a click and hope it gets submitted on the first try thing.

nEEtdo0d:
So in particular Cheshire, Champagne, and Hakyru are examples of taking almost nearly all their grid space so sharing a common spot for `Navbar` wouldn't quite work.

nEEtdo0d:
Drafted but mixed results because I hadn't noticed before but even the rightmost index color changes indefinitely depending on who is the focused ship because the area is semi-transparent. Sometimes getting a total of 5 rather than 6. And as expected, using the same `Navbar` for a different series can either go well or terrible haha. So a consideration is a personal `Navbar` for each series but Champagne or Hakyru really doesn't have a common area, Cheshire seemed okay though.

https://pastebin.com/KW3kDq4e

Something I hadn't noticed before, the ship focused is dependent on who is currently being researched, if the series changes the location is the same, so this can be problematic as well. I think if no ship is being researched then it just defaults to the most leftmost.

LmeSzinc:
Use the entire grid, it doesn't need to be a corner.
```
SHIPYARD_FACE_GRID = ButtonGrid(origin=(188, 607), delta=(181, 0),
                                       button_shape=(181, 80), grid_shape=(6, 1),
                                       name='SHIPYARD_FACE_GRID')

shipyard_bottom_navbar = Navbar(grids=SHIPYARD_FACE_GRID,
                  active_color=(33, 93, 173), active_threshold=221, active_count=50,
                  inactive_color=(66, 73, 99), inactive_threshold=221, inactive_count=50)

logger.info(shipyard_bottom_navbar.get_active(main=self))
logger.info(shipyard_bottom_navbar.get_total(main=self))
shipyard_bottom_navbar.set(main=self, left=1)
logger.info(shipyard_bottom_navbar.get_active(main=self))
logger.info(shipyard_bottom_navbar.get_total(main=self))
```

nEEtdo0d:
> Use the entire grid, it doesn't need to be a corner.

Good call, wouldn't have considered it was too focused on the common area bit. 

You are right by expanding grid area, increasing the threshold to ignore the myriad of colors the image might have besides the active/inactive color, and making the count relatively small seems to do the trick even for the ones that basically take up 90% of the grid.  The `origin` update is appreciative too, initially started with what I had but just having that extra pixel drastically changed the detection success rate.

Did an isolated test to step through all series and randomly pick an index to see whether it was consistent regardless of where the active index was, it would repeat that from its present location.  It passed impressively not having detected more than 1 active state.

PR #561, opened 2021-08-07
Add: support to run reward loop during OS

nEEtdo0d:
Pretty straight forward, check and runs reward loop before going for a zone clear. Been using this myself for the past few monthly resets.

I'll keep the PR open for a few days to gauge whether this is the direction to go with or preferably use a separate instance rather than inheritance like
```
import Reward
backup = ...
check = Reward(...)
if check.handle_reward():
    ....
backup.recover()
```

If no particular backlash, I can make the merge official.


nEEtdo0d:
RIP, because of shop module inherited by Reward I used the same method signature, this overrides the os_shop methods. So have to use the standalone instance method.

LmeSzinc:
You can rename methods in module/os_handler/shop.py to `os_shop_xxx()`, and rename `ShopHandler` to `OSShopHandler`

nEEtdo0d:
> You can rename methods in module/os_handler/shop.py to `os_shop_xxx()`, and rename `ShopHandler` to `OSShopHandler`

Gotcha then we can maintain the overall direction in that case. Didn't want to overstep, thanks.

nEEtdo0d:
Pushed directly to master to minimize commit history

Issue #564, opened 2021-08-08
module/shipyard/ui.py 第80行 _shipyard_get_bp_count 方法返回剩余蓝图数错误


railzy:
```python
    def _shipyard_get_bp_count(self, index=0):
        """
        Args:
            index (int): Target index's BP count

        Returns:
            Ocr'ed count for index
        """
        # Base Case
        if index < 0 or index >= len(SHIPYARD_BP_COUNT_GRID.buttons):
            logger.warning(f'Cannot parse for count from index {index}')
            return -1

        result = OCR_SHIPYARD_BP_COUNT_GRID.ocr(self.device.image)

        return result[index]
```

由于index是从1开始数船所处位置顺序，而python列表是从0开始索引的，这里返回了所指定船位置+1处的船的蓝图数，导致错误识别蓝图数量，且可能导致索引越界

应该这样才对
```python
return result[index-1]
```

nEEtdo0d:
Thank you, I had overlooked this after bumping up the index after converting over to `Navbar`

PR #565, opened 2021-08-08
Fix: return right BP count of target index

railzy:
#564


nEEtdo0d:
Thanks, I overlooked this after adjusting the index input. You are correct, incorrect index

PR #566, opened 2021-08-08
Fix: [JP] assets for new module

railzy:
assets for new module 

I am not sure what is `DOCK_SELECTED` so I haven't updated it

I haven't get the screenshots of `RESEARCH_COMPLETE` and `SHOP_SELECT_PR*` so they are not updated


LmeSzinc:
1. DOCK_SELECTED is :
2021-08-09_18-46-34_aYk0BBdgju_924_520

2. SHOP_SELECT_PR_* should be the same among servers.
3. About guild shop, please choose to follow EN or CN, edit `module/shop/guild_shop.py` `GuildShop.shop_guild_items()`.
Items in EN are smaller.
guild_shop_cn_cn_diff

railzy:
It seems that assets of guild shop items in JP are the same as EN. 
ALAS can check and buy guild shop items correctly in JP server without any change now.

PR #568, opened 2021-08-11
Fix: [TW] Shop module asset

cola228:
The size of the picture in guild shop is the same as the picture in CN , thank you.


LmeSzinc:
Copy the CN version of `shop_guild_items()` and set `SERVER='tw'` in decorator.

PR #579, opened 2021-08-16
Make it more stable

Cyumxi:


Cyumxi:
Rewrite all while loop as self.ui_click

PR #580, opened 2021-08-16
Fix: Collect weekly mission rewards

nEEtdo0d:
Helps avoid timeout for when entering in `page_mission` because of red dot on `page_main` but only weekly missions are available to collect which aren't accessible on the `all` navbar page

Gauging whether going in the right direction with code reuse, add a new clause, adjust conditionals to handle the both the `weekly` and `all` navbar pages for certain assets, and some re-arranging.

Ran after piling up on missions for this week and all gone, so would have to wait another week to test it from scratch again.  But seems stable from this recent run through


module/reward/reward.py
```diff
@@ -186,29 +193,37 @@ def _reward_mission(self):
...
                 click_timer.reset()
+                timeout.reset()
+                continue
+
+            if self.handle_popup_confirm('REWARD_MISSION'):
                 exit_timer.reset()
+                click_timer.reset()
                 timeout.reset()
+                reward = True
                 continue
```
LmeSzinc, line 227: Since the "in" and the "out" of `_reward_mission_collect()` are both page_mission, variable `reward` is used to avoid exiting before receiving any missions. It should be set to True only in get_items, and should be set to False at the start of `_reward_mission_collect()`

nEEtdo0d: Right the original had `reward = True` only under `GET_ITEMS` clause but in the cases the only collection taking place are certain rewards which do not result in `GET_ITEMS` appear, it would time out instead since it would not set the flag so it seemed the only option was to add it to different clauses too.  

According to [**wiki**](https://azurlane.koumakan.jp/Task), the 250, 500, and 800 point ladder.  Specialized core when capped for the month would result in `POPUP_CONFIRM` and gold bulin being the only reward just has `GET_SHIP` no `GET_ITEMS`  in either of these cases so results in exit by time out.  But if you think its safer this way and have only one clause act as `reward = True` I can remove the others.

Likewise, since I was dealing with `reward` across two pages, I had to consider carrying its value over in the case weekly was the only thing being collected, time out under the all page.  But I can remove the parameter altogether and have `reward` be just a local variable in `_reward_mission_collect`.  Two time outs may occur in sequence in the worst case but still better than entering the `page_mission` every time but with no way to collect weekly.

The `interval` parameter is needed according to tests, having it set to `1` for weekly, resulted in early termination and sometimes actually stuck in `GET_ITEMS` resulting in failure altogether. `0` was the only argument that allowed it to collect without error even `0.5` surprisingly didn't work.

The primary goal is to collect weekly missions and not leave ALAS constantly enter `page_mission`.  I hoped we could also reduce the need to time out but yes there is the risk of exiting too early with `reward = True` being in several clauses though I've yet to come across that situation as of yet.

Let me know if this is preferred, I can make the appropriate updates with that.

nEEtdo0d:
Did a little re-work in a different branch and has a PR open, will close whichever is preferred. 

https://github.com/LmeSzinc/AzurLaneAutoScript/pull/583

PR #632, opened 2021-09-27
Add: Daily Gacha Reward

nEEtdo0d:
PR since LmeSzinc is working on a big infrastructure change in branch `app` so wasn't sure whether this could be merged smoothly into there being the old implementation style.

Automate the gacha process, some cases result in `exit` not always able to handle those gracefully so needs degree of user intervention to continue.


LmeSzinc (commented):
I think it's unneccessary to add wishing will pull. It only opens a few times a year. Game UI might be changed and we need more time to maintain it.

nEEtdo0d:
> I think it's unneccessary to add wishing will pull. It only opens a few times a year. Game UI might be changed and we need more time to maintain it.

Well personally I'm missing ships from the pools so find this needed but I can remove the option in `argparser`.  It uses the `navbar.get_total` like if `event` is available so maintenance wise it doesn't really differ  The assets cut are from old source images but I think still match to last wishing well.

> Use `handle_popup_confirm()` here.

I would have but it uses the `POPUP_CANCEL` in its check, the cancel asset in this is red vs typically it would be gray

> Call if self.handle_retirement(): to handle dock full

Good catch, I was under the impression when emptying the queue it stayed in the `page_build` no matter how many clicks but it does indeed display the window if dock is full so the behavior I have isn't needed then.

Thanks

LmeSzinc:
I was hoping to remove BUILD_SUBMIT_WW_COUNT, BUILD_SUBMIT_WW_ORDERS, BUILD_WW_CHECK and their use.
Add the options back if it's neccessary to use those assets.

PR #653, opened 2021-10-15
Add: Reward task DataKey

nEEtdo0d:
I can move the module to its own folder like others have but didn't want to re-run the button_extract if this idea was going to be rejected and I imagine there will be more edits needed anyway with a part I couldn't figure out. However it does seem to be functioning and running successfully.

The part I couldn't figure out was how to update the i18n directory with the language JSON files, tried updating them but then the GUI launched but no menu/options were displayed. It didn't crash the GUI and no console output to view so I couldn't tell what was wrong.

As soon as I put it back to its head revision, it worked fine and still had the DataKey option available to be picked and configured.

There was also the option of just integrating into the existing war archives but it running again in case it resumes a campaign seemed like an inefficient idea so even though this task is pretty bare bones, still a better to do so than tag it along to another task.

### DIFF
--- a/module/config/i18n/en-US.json
+++ b/module/config/i18n/en-US.json
@@ -111,8 +111,8 @@
       "help": ""
     },
     "DataKey": {
-      "name": "Task.DataKey.name",
-      "help": "Task.DataKey.help"
+      "name": "DataKey",
+      "help": ""
     },
     "Daily": {
       "name": "Daily",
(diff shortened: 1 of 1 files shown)


LmeSzinc:
Run `python -m module.config.config_updater`, everything will be auto generated

nEEtdo0d:
> Run `python -m module.config.config_updater`, everything will be auto generated. The end of config_updater.py simply describes the auto-update process

Great, works now and was figured out how this translate option works. Might do more at a later time. But otherwise branch looks good for merge.

PR #772, opened 2021-11-25
Fix: Refactor RewardTacticalClass

nEEtdo0d:
Fixes:
- re-factor tactical exp progress check
- streamline book removal process
- if resort to 'first' not a Book type, so no button attribute

### DIFF
--- a/module/tactical/tactical_class.py
+++ b/module/tactical/tactical_class.py
@@ -185,17 +185,26 @@ class RewardTacticalClass(UI):
         logger.warning('No book found.')
         raise ScriptError('No book found, after 15 attempts.')
 
-    def _tactical_selected_get(self):
+    def _tactical_book_select(self, book, skip_first_screenshot=True):
         """
-        Find the selected book onscreen
-        Called after _tactical_books_get
+        Select the target book onscreen
+        Updates current image if needed
+
+        Args:
+            book (Book):
+            skip_first_screenshot (bool):
         """
-        selected = None
-        for book in self.books:
-            if book.check_selected(self.device.image):
-                selected = book
+        while 1:
+            if skip_first_screenshot:
+                skip_first_screenshot = False
+            else:
+                self.device.screenshot()
+
+            if not book.check_selected(self.device.image):
+                self.device.click(book.button)
+                self.device.sleep((0.3, 0.5))
+            else:
                 break
-        return selected
 
     def _tactical_books_filter_exp(self):
         """
@@ -203,16 +212,35 @@ class RewardTacticalClass(UI):
         books from self.books based on current
         progress of the tactical skill.
         """
-        selected = self._tactical_selected_get()
-        if selected is None:
-            return
+        # Shorthand referencing
+        first, last = self.books[0], self.books[-1]
 
+        # Read 'current' and 'remain' will be inaccurate
+        # since first exp_value is factored into it
         current, remain, total = SKILL_EXP.ocr(self.device.image)
+
+        # Max level in progress; so selective books
+        # should be removed to prevent waste
         if total == 5800:
-            # Read 'current' and 'remain' are inaccurate
-            # as selected exp_value is factored into it
-            current -= selected.exp_value
-            remain += selected.exp_value
+            if current == 0:
+                # Lvl 9+1, using first will reach max level
+                # Swap to last and re-OCR
+                self._tactical_book_select(last)
+                current, remain, total = SKILL_EXP.ocr(self.device.image)
+                if current == 0:
+                    # Still Lvl 9+1 even with last
+                    # Must re-calculate to accurately gauge
+                    current = total - last.exp_value
+                    remain = last.exp_value
+                else:
+                    # Lvl 9, so can calculate normally
+                    # but use last
+                    current -= last.exp_value
+                    remain += last.exp_value
+            else:
+                # Lvl 9, so can calculate normally
+                current -= first.exp_value
+                remain += first.exp_value
             logger.info('About to reach level 10; will remove '
                         'detected books based on actual '
                        f'progress: {current}/{total}; {remain}')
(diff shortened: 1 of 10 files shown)


LmeSzinc:
I noticed it's timewasting when you have EXP in 4700/5800 and ALAS removed the T3 books, so I would like to introduce new a setting group. New options also remind users there is such a feature in ALAS.

**EXP Overflow Settings**
- Control EXP overflow
- Allow to waste X EXP of a T4 book
- ...T3
- ...T2
- ...T1

Internal names will be:
```
ControlExpOverflow:
    Enable: true
    T4: 100
    T3: 100
    T2: 200
    T1: 200
```

nEEtdo0d:
> I noticed it's timewasting when you have EXP in 4700/5800 and ALAS removed the T3 books, so I would like to introduce new a setting group.

Ran a couple tests using a fixed image and changing the `remain` for different scenarios.

Seems to behave like it should whether enable is set to `True` or `False`

> New options also remind users there is such a feature in ALAS.

Sorry not sure what you mean by this, is there an area where I should record new features/optimizations that display or remind to users?

module/tactical/tactical_class.py
```diff
@@ -175,6 +185,72 @@ def _tactical_books_get(self, skip_first_screenshot=True):
...
+                # nothing else
+                if remain < (exp_value * factor):
+                    groups.append(self.books.select(tier=tier, exp=True))
+                if exp_value != 100 and remain < exp_value:
+                    groups.append(self.books.select(tier=tier, exp=False))
+                logger.info(f'T{tier} removing: {sum([group.count for group in groups])}')
+
+                # Remove applicable books to prevent waste
+                for group in groups:
+                    self.books = self.books.delete(group)
```
LmeSzinc, line 252: Each `Book` object has attribute `exp_value`, no need to calculate again.
A simple implement would be:
```python
def filter_func(book):
    if current + book.exp_value > total + self.config.ControlExpOverflow_T*Allow:
        return False
    return True

self.books = SelectedGrids([b for b in self.books if filter_func(b)])
```

nEEtdo0d: Thanks for the sample, quite elegant. I'll run simulations to ensure its behaving as intended. From what it looks like we don't need the `ControlExpOverflow_Enable` then, people can just put in zero if they rather not have any overflow for whatever tier.

Separate issue; when going from level 8 to 9, the OCR will read a digit count X/5800, that makes keeping track of `current` more difficult. 
```
2021-11-26 11:15:01.999 | INFO | [OCR_SKILL_EXP 0.082s] 0/5800
2021-11-26 11:15:02.000 | INFO | About to reach level 10; will remove detected books based on actual progress: -3000/5800; 8800
```
Thinking of capturing an asset where it might say lv 8+1, might be a while since I don't have any at the moment.  Using lv 9 asset + 5800 might work too...

Now that I think about it, also might have been read as `0/5800` because could print as `MAX/5800`. Hmm, then wouldn't be able to gauge an accurate `current` can't assume you didn't overflow already in this case and wouldn't know how much overflowed. Guess I can see the reason this filter being more difficult than I anticipated.

Still managed process fine though for now.

LmeSzinc: Try this pseudo code:
```
(do OCR)
if current == 0:
    # Lv9+1, using the first book will level to max
    (select the last book)
    (do OCR)
    if current == 0:
        # Lv9+1, using the book with minimum exp will still level to max
        (current = total - selected_book.exp_value)
    else:
        # Lv9, current < total, able to get actual current exp
        (current -= selected_book.exp_value)
else:
    # Lv9, all good, able to calculate
    (current -= selected_book.exp_value)
```

nEEtdo0d: Definitely was overthinking and I only just realized I was mistaken about the `Lv8+1`, it does not change `total` dynamically like I had thought.  So only have to worry about 1 case which your pseudo code did simplify a whole lot.

Did a live test since using a fixed image wouldn't have worked.  Fortunately had one skill close to max level and seems to work
```
2021-11-27 12:05:27.895 | INFO | [Book_count] 12
2021-11-27 12:05:27.895 | INFO | [Books] [Red_T4_Exp, Blue_T4, Yellow_T4, Red_T3
_Exp, Blue_T3, Yellow_T3, Red_T2_Exp, Blue_T2, Yellow_T2, Red_T1_Exp, Blue_T1, Y
ellow_T1]
screenshot: 0.4775776863 s
2021-11-27 12:05:28.380 | INFO | [Book_count] 12
2021-11-27 12:05:28.380 | INFO | [Books] [Red_T4_Exp, Blue_T4, Yellow_T4, Red_T3
_Exp, Blue_T3, Yellow_T3, Red_T2_Exp, Blue_T2, Yellow_T2, Red_T1_Exp, Blue_T1, Y
ellow_T1]
2021-11-27 12:05:28.381 | INFO | Click ( 280,  326) @ BOOKS_GRID_0_0
2021-11-27 12:05:28.381 | INFO | <<< MINITOUCH INIT >>>
2021-11-27 12:05:30.383 | INFO | Minitouch bind to port 20793
2021-11-27 12:05:30.513 | INFO | Minitouch forward port 20793
2021-11-27 12:05:30.517 | INFO | minitouch running on port: 20793, pid: 5185
2021-11-27 12:05:30.517 | INFO | max_contact: 2; max_x: 32767; max_y: 32767; max
_pressure: 0
screenshot: 0.5068545341 s
2021-11-27 12:05:31.472 | INFO | Loading OCR model: ./bin/cnocr_models/azur_lane

2021-11-27 12:05:31.549 | INFO | [OCR_SKILL_EXP 0.090s] 0/5800
2021-11-27 12:05:31.549 | INFO | Click ( 979,  483) @ BOOKS_GRID_5_1
screenshot: 0.5137686729 s
2021-11-27 12:05:32.543 | INFO | [OCR_SKILL_EXP 0.023s] 5450/5800
2021-11-27 12:05:32.543 | INFO | About to reach level 10; will remove detected b
ooks based on actual progress: 5350/5800; 450
2021-11-27 12:05:32.544 | INFO | [EXP Filtered] 6
2021-11-27 12:05:32.544 | INFO | [Books] [Red_T2_Exp, Blue_T2, Yellow_T2, Red_T1
_Exp, Blue_T1, Yellow_T1]
2021-11-27 12:05:32.544 | INFO | [Book_sort] Red_T2_Exp > Red_T1_Exp > Blue_T2 >
 Yellow_T2 > Blue_T1 > Yellow_T1 > first
2021-11-27 12:05:32.545 | INFO | Click ( 286,  475) @ BOOKS_GRID_0_1.
```

PR #837, opened 2021-12-17
Fix: add filter set before retirement

zsx2034:


zsx2034:
The different rarity name (eample, N with common) maybe a little bit bad.

LmeSzinc:
Call `dock_filter_set()` without arguments to reset dock filters after retirement.

zsx2034:
I added it just now.

PR #850, opened 2021-12-24
Add: re-factor shop module and limited core shop support

nEEtdo0d:
Re-factor the shop module using key methods like others funcs; compress shop_buy_execute to one py file, customize via optional func definitions specific to shop type

Limited to just purchasing chips and arrays. Probably for most players it would just be arrays as buying chips is easily accruable through other sources


LmeSzinc:
This will require a lot of work, but is it possible to  call functions and use configs without `getattribute`?
Using spliced strings to do that brings great difficulties into tracking and reading the code.

nEEtdo0d:
> This will require a lot of work, but is it possible to call functions and use configs without `getattribute`? Using spliced strings to do that brings great difficulties into tracking and reading the code.

Probably but as you said, I think we would need to start from the ground up for this endeavor.

My opinion might be biased seeing as the module was primarily authored by me, but I guess I don't have that much difficultly following the infrastructure and reading because of it.  Initially using the `os_shop` design as a base which dealt with only two shops and were functionally the same.  This variant had to deal with several others with their interfaces being slightly different and some even having different buy handling events (guild and core in this case). 

So it made sense to use a spliced string methodology to differentiate them and avoid placing them all into a single py file.  Also I personally didn't like the idea of having `if-else` clauses everywhere calling their respective function when the spliced string could compress them.

I liked the idea and it was pretty minimal at first.  But overtime definitely getting unwieldy with the new extra layers and needing to pass the value almost constantly.  It might be worth just making it an internal attribute (in `shop_buy` then do `self._shop_str = shop_type` and then plugging into the various function calls directly.  No need to keep carrying from function to function as we are now.  We wouldn't necessarily need to call `getattribute` as we can attempt to call it directly as well and we'd still use try/catch blocks to avoid exceptions.

`self.shop_x_y() == self.__getattribute__(f'shop_x_y')()` behave the same way and throws the same exception.

The infrastructure wouldn't have really changed but at least less passing around and just figuring it doesn't change during execution anyway might as well make it locally available at all times.

Currently shifted focus on a different mechanism, meowfficer feeding (the first half is going well, the second half is kinda iffy not looking forward to it haha), also being holidays so for now would like to suggest merging as it is.  Revisit the idea later as it does function stably from what I've tested thus far.

LmeSzinc:
A quick migrate could be: (pesudo code, not runable)
```python
class ShopBase(UI):
    def select_items(self):
        return []

    def buy(self):
        for item in self.select_items():
            pass


class GeneralShop(ShopBase):
    def select_items(self):
        return ['one_item', 'another_item']


class MeritShop(ShopBase):
    def select_items(self):
        return ['one_item', 'another_item']


class RewardShop(UI):
    def run(self):
        GeneralShop(config=self.config, device=self.device).run()
        MeritShop(config=self.config, device=self.device).run()
```

nEEtdo0d:
> A quick migrate could be: (pseudo code, not runnable)

I think this might work if I am understanding it right.  Stepping away from using a grand inheritance scheme and turn to having the a divided subtask like pattern instead. I'll start a new branch in the near future to try out this forward thinking.  Thanks for the basis!

Going to merge this as is for now to get core shop support as it will be refreshing tomorrow.

Thanks

PR #861, opened 2022-01-01
Add: meowfficer enhance support task

nEEtdo0d:
Wouldn't say its less complex than the last PR that tried to tackle this but a different approach in trying to minimize the number of screenshots for certain actions.

Also limited to just running daily like all the other meowfficer tasks.  I would think burning 10 a day is sufficient for most people at least is for me.  Although there is that speed finish ticket that may be worth pursuing later.

Helps burn through boxes using a different queue method to ensure the common boxes are prioritized in opening rather than using the quick button.  Still would want to keep the SR cats in case of unique skills I would imagine


yanyihui:
I closed my last PR because when I communicated with lmeszinc, he mentioned the need to feed the moewfficer through index. It means you need to scan more than one page of the meowfficer. 12 is not enough.
In issue #852 , I suggest that the scroll class is not well compatible with moewfficer pages. I can't solve the problem of de duplication of list reading. 
Maybe using screener is a better way. If u want to try it, see my new pr. I will commit en page soon.

nEEtdo0d:
> I closed my last PR because when I communicated with lmeszinc, he mentioned the need to feed the moewfficer through index. It means you need to scan more than one page of the meowfficer. 12 is not enough. In issue #852 , I suggest that the scroll class is not well compatible with moewfficer pages. I can't solve the problem of de duplication of list reading. Maybe using screener is a better way. If u want to try it, see my new pr. I will commit en page soon.

Oh I see, thanks for keeping me posted.  This is true, a limitation of this can only select the first 12. But I came to the conclusion that even with this limitation, the goal really was to burn through excess boxes and the amount of XP you get is mostly negligible and it would take a very long time for someone to reach the max level for all 12 slots.

Yes the scrolling adds a lot of complexity so I had avoided this myself haha.  Especially with sensitive click locations and I can imagine there would also be the problem of duplication.  The commission module I think might be a good reference but its a bit too unwieldy for me to read.  It sort of deals with a similar issue but somehow works.

The alternative approach would be to use the filter and sorting mechanism very similar to the retire module.  Somewhat expands from the limited 12 and would involve multiple configs I think.  But since I wanted to keep it to one config, I opted not pursuing that path.

Happy New Year btw!

yanyihui:
Thank u for your wishes.

I'm afraid my need is not just to consume excess boxes. I don't know the English expression of meofficers' skills. Maybe more people want to have specific skills. Like fire(damage caused increased by 3%) and mountain(reduce damage received by 3%). When u get a new meowfficer which has these skills. You may want to feed it. Reseting some useless skills when it grow up to level 15.

I used to read an article. It is mentioned that the optimal level for resetting skill points is level 15 to 20. Higher levels are hard to reach. Lower levels have fewer skill points. Under level 15, the experience gained from the box is quite considerable(3 skill points for sr = 59900 exp, 200 boxes). 

But the number of config you raised is really a problem. I don't know if others have the same need as me. Maybe more people don't have many coins to buy boxes. However, as a user of alas, I think coins is not a problem( :) )

LmeSzinc (commented):
Feeding meowfficers requires a little coins, it's better the add a simple checker like stop feeding if coins < 1000. You can start collecting images of special skills from now. Locking gold cats doesn't seem to be  a good idea, because only skills that matters. Alas should only lock those with at least one special skills in the future.

Some other things about the meowfficer module, since the the lines of meowfficer.py is increasing, I would like to split it into base.py, buy.py, chore.py, train.py, feed.py, and meowfficer.py which simply calls then and does the schedulings.

nEEtdo0d:
Back to tackling. Got most of your concerns and still in the planning/imaginative/drafting phase for others.

- Box count OCR often incorrectly read or defaults to zero due to failed reading.  Definitely was fine when I had 100s of boxes but now after a month of using, not doing so hot after getting to low 40s. May consider read in main screen and maintain count internally.

- parse meowfficer skills; this one may be difficult
-- The images of skills are more or less the same visually even the unique ones.  But particularly, the only difference are the numerals in the bottom left.  Matching might not go so well but I'll will try first
-- When collecting screen will randomly shift left. This is why there are some assets with `offset=(40, 20)`.  Doing match template for these situations will definitely throw off the `grid` definition since its parameters are fixed.
-- Collect will be slower as waiting to analyze skills, applying to all rarities. Maybe capture assets to discern rarity collected; most cases you only want unique skills when it is an SR I would presume. 
-- Likely will add a new func `handle_meow_popup_confirm/cancel` variant. May try to utilize lock button to simplify check looping.

- Break up the module into pieces; question would you prefer it be an inherited design or like `shop` no inheritance and run each? Tackle last, easier to split when you know everything already works as is and just needs re-arranging anyway.

LmeSzinc:
> question would you prefer it be an inheritance design or like shop no inheritance and run each separately?

The inheritance design.

To be clear: lock SR/SSR meowfficers with at least 1 unique skill only. Excluding all R meowfficers is a easy way to detect rarity, since they all share the same appearance. About detecting the skills, try cropping the bottom left area.

nEEtdo0d:
#923 serves as enhanced version of this PR in addition to refactoring the meowfficer module

PR #891, opened 2022-01-20
Opt: shop revamp final

nEEtdo0d:
Stable as of a few test runs

Sorry for the delay and restructure things took a while. But pretty worth the trouble, no string passing, streamline parsing with regexp, and different approach running as individual components rather than single entity

When merged, will need to announce and change the documentation on the wiki regarding shop filter string has changed slightly to adopt that regexp and filter class usage
`PRBP -> PR`, `PRNeptuneBP -> PRNeptuneS1 or PRNeptune`, `PR1BP -> PRS1` likewise changes for `DR`


module/config/config_generated.py
```diff
@@ -203,7 +203,7 @@ class GeneratedConfig:
 
     # Group `GuildShop`
     GuildShop_Refresh = True
-    GuildShop_Filter = 'PlateT4 > BookT3 > PRBP > CatT3 > Chip > BookT2 > Retrofit > FoodT6 > FoodT5 > CatT2 > BoxT4'
+    GuildShop_Filter = 'PlateT4 > BookT3 > PR > CatT3 > Chip > BookT2 > Retrofit > FoodT6 > FoodT5 > CatT2 > BoxT4'
```
LmeSzinc, line 206: Don't chenge this file manully, edit `module/config/argument/argument.yaml` and run `config_updater.py`

LmeSzinc (commented):
ConfigUpdater also has a redirection list which can update user's filters instead of making an announcement and telling users to do that. Use regex to replace the old with the new.

module/config/config_generated.py
nEEtdo0d: Understood sorry about that.  I did a global search on only PY files hadn't realized YAML too.  But after update I ran the module, no change to them since it did output the same. 

Wouldn't users still have to update it manually if they made their own filter or because the old one is saved in their JSON?

LmeSzinc: > Wouldn't users still have to update it manually if they made their own filter or because the old one is saved in their JSON?

ConfigUpdater also has a redirection list which can update user's filters.
Add a redirection `('ShopOnce.GuildShop.Filter', 'ShopOnce.GuildShop.Filter', func)`, func is a function that receives the old value and returns the new value. Use regex to replace `PROdinBP` with `PROdinS3` for example.

nEEtdo0d: Oh I see I wasn't sure what you meant by redirection. But I'm not sure how to use this effectively.  How would I plug in a `func` into this since at this scope I wouldn't have `self`? Do I use `lambda`? I think even with regex I wouldn't be able to compress this very easily. 
Rough code
```
def shop_convert(value):
    name_to_series = {neptune: 1, ..., odin: 3}
    str = '^(pr|dr)(neptune|...|odin)(bp)$'
    regex = re.compile(str, re.IGNORECASE)
    result = re.search(regex, value)
    if result:
         groups = result.groups()
         if groups[1] is not None:
              # Build equivalent string PROdinBP to PROdinS3, use dict to correlate name to series
              return f'{groups[0]}{groups[1]}S{name_to_series [groups[1]]}'
         else:
             # No subgenre so safe to assume is PRBP which converts to PR
             return f'{groups[0]}'

    # Not an old PR/DR filter value so return as is
    return value
```

nEEtdo0d: Just realized the value being fed into this is the entire string so that pseudo code wouldn't be quite accurate.  The conditions might still be relevant.

`re.sub(regex, f'\1\2\3', value)` is probably closer to what you had in mind.  But I think it will be complicated. Can I add conditional logic into an expression and also use the captured group notation at the same time. Like `{name_to_series[\2]}`?

`name_to_series = {Neptune: (Neptune, 1), ...` the dict becomes a little more complicated and repetitive, lack of common ground between `PR1BP to PRS1` and `PRNeptuneBP to PRNeptuneS1`

I guess I could just brute force it, having the dict `{1: S1, Neptune: NeptuneS1}` so be easier to sub in and not worry about the different format, not having to worry about the `\3` captured group at all.

LmeSzinc: No problem with brute force, update it silently in the way you like.

nEEtdo0d: Alright best I could come up with that seems to do the job.
[shop_filter_redirect_utils.txt](https://github.com/LmeSzinc/AzurLaneAutoScript/files/7916117/shop_filter_redirect_utils.txt)

So next plug in?
```
from module.config.shop_filter_redirect_utils import ship_bp_redirect
redirection = [ ..., ('ShopOnce.ShopGuild.Filter', 'ShopOnce.ShopGuild.Filter', ship_bp_redirect), ...]
```

Keep in config module or should be in new one to isolate all `redirect_utils` in one place? Was imagining the infrastructure to be multiple `redirect_utils` files based grouped by their respective config so it wouldn't end up just as 1 giant file.

LmeSzinc: Put them in `redirect_utils`

nEEtdo0d: Successful test run redirecting the configs to the new format. Still should probably update the wiki not sure if I have write access to there.  But if there is nothing else, I can merge this optimization. Let me know, thanks.

LmeSzinc: Everything is good and can be merged. Collaborators have the privilidge to edit wiki, do when you have time.

PR #948, opened 2022-03-05
Upd: meowfficer re-work allow retention of gold and purple meowfficer

nEEtdo0d:
Configurable to retain gold and/or purple meowfficer if has a special talented ability. Capable of retaining neither as well.

Keep in mind:
- Due to rarity tier differences, servers will need to cut their own counterpart asset for gold and purple (EN=SR, Elite, CN=SSR, SR)
- Due to the lock and unlock icon being very similar in appearance which Alas can't tell the difference, using text to differentiate was better. However you may need to be precise in the cut due to the random behavior of the screen shifting.  As a result both the icon and text are partially cut from the screen in this case, so asset need to be partial part of the word as well. 
-- Sample images: https://imgur.com/a/rN1tJxp


LmeSzinc:
Thanks for the info, will do that.

PR #1006, opened 2022-04-11
Fix: Adding semantics to str input of num type

hgjazhgj:
Yesterday, the serial number `27e834120906` of one of my phones has been interpreted and stored as a float, and alas crashed.  
I think it's stupid to decide on data types based on literal values alone without looking at the actual semantics.  
Notice that some items in args have `validate` field to ~~determine it to datetime~~ reject invalid input.  
So I use this field to tell int/float while items without this field will keep as a string.  
This pr makes it necessary to specify `validate` field as int/float for numeric inputs added later.  
Btw, only non-negative inputs are accepted now.

### DIFF
--- a/module/config/argument/args.json
+++ b/module/config/argument/args.json
@@ -196,11 +196,13 @@
       },
       "SuccessInterval": {
         "type": "disable",
-        "value": 0
+        "value": 0,
+        "validate": "int"
       },
       "FailureInterval": {
         "type": "disable",
-        "value": 0
+        "value": 0,
+        "validate": "int"
       },
       "ServerUpdate": {
         "type": "disable",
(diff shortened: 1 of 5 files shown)


18870 (approved):

module/config/argument/args.json
```diff
@@ -4466,11 +4606,13 @@
...
-        "value": "60-120"
+        "value": "60-120",
+        "validate": "int"
       },
       "FailureInterval": {
         "type": "disable",
-        "value": "60-120"
+        "value": "60-120",
+        "validate": "int"
       },
```
LmeSzinc, line 4616: 这里不知道会不会有问题

LmeSzinc (commented):
~~都是中国人就不讲英文了~~

module/config/argument/args.json
```diff
@@ -6357,7 +6572,8 @@
...
       "TargetZone": {
         "type": "input",
-        "value": 0
+        "value": 0,
+        "validate": "int"
       }
```
LmeSzinc, line 6577: TargetZone 应该同时允许 int 和 str

module/config/argument/args.json
hgjazhgj: 理论上,这个type是disable,不会有需要校验的输入,所以这个修改不会导致报错  
但是显然这里摆个validate不合适,我考虑下怎么改,比如不对disable提供validate字段

PR #1155, opened 2022-05-27
Fix: encountered bugs, guild shop issues

nEEtdo0d:
- Meowfficer feed material noticeably lags on first boot up of emulator or after collecting many trained meowfficers
- Enhance phase can still be re-entered, match the conditions for both to ensure exit
- Handle issues from #1133 and #990; optimize implementation and handle odd bug in stock


LmeSzinc:
Imitate `SkillExp`, this should recognise all red/green/white letters.
```python
class SkillExp(DigitCounter):
    def pre_process(self, image):
        # Image is like `NEXT:1900/5800`, 1900 is green and others are in white
        # Convert to gray scale
        r, g, b = cv2.split(image)
        image = cv2.max(cv2.max(r, g), b)
```

nEEtdo0d:
My bad sorry saw the checks passed and merged but I can re-visit it if my approach isn't to your liking.

Good idea to use gray scale, didn't know or think I could do that.

Worked around it so didn't need gray scale.  The `Digit` ocr filters out the white characters in that number, grabbing only the green numbers.  Also just checking for red pixels to get past the overbuying circumstance. 

Though if I were to use gray scale, could potentially simplify to something like this?
```
current, remain, total = ocr_stock.ocr(image)
if remain == limit or not current: 
    return limit
return ocr_center.ocr(image)
``` 

instead of using `image_color_count`

nEEtdo0d:
My bad sorry saw the checks passed and merged but I can re-visit it if my approach isn't to your liking.

Good idea to use gray scale, didn't know or think I could do that.

Worked around it so didn't need gray scale.  The `Digit` ocr filters out the white characters in that number, grabbing only the green numbers.  Also just checking for red pixels to get past the overbuying circumstance. 

Though if I were to use gray scale, could potentially simplify to something like this?
```
current, remain, total = ocr_stock.ocr(image)
if remain == limit or not current: 
    return limit
return ocr_center.ocr(image)
``` 

instead of using `image_color_count`

PR #1179, opened 2022-06-03
Full and uncompromising Python reformatting

alessio-ds:
I've done a full and uncompromising Python code reformatting using [Black](https://pypi.org/project/black/).

**This doesn't affect the Python codes in any way** - it simply makes them more readable and easily reviewable.


campaign/campaign_main/campaign_10_2.py
```diff
@@ -24,44 +24,90 @@
...
+    H5,
+    A6,
+    B6,
+    C6,
+    D6,
+    E6,
+    F6,
+    G6,
+    H6,
+) = MAP.flatten()
```
LmeSzinc, line 84: I don't think you have read the development docs or any lines of code in ALAS. If you did, you'll know none of these changes will improve readability.

LmeSzinc (commented):
There're too many similar PRs in your account history, you are not here to contribute open source projects.

campaign/campaign_main/campaign_10_2.py
alessio-ds: The process of reformatting is fully automatic, sorry if I didn't check every single file.
Overall, except this problem with the readability of the campaign levels, to me the code looks better (better indents, string management, variables and so on).
Other than being my personal opinion, remember that Black is used at an enterprise level.

I would consider leaving unchanged the campaign .py files, and merge the rest.
I understand that this is a 955 files changed commit, but the reason I did this is because I was surprised you didn't use Black before.

This said, don't judge people based on PRs history. I wouldn't be here if I didn't want to give a hand.

PR #1237, opened 2022-06-20
Add ascreencap for android 8 (oreo) and 9 (pie)

n0k0m3:
Currently ascreencap is only available for android 5-7. This PR will add ascreencap for 8 and 9 (from ClnViewer repo).

There's a major rewrite in Android 10 and 12 for SurfaceComposerClient so we'll need to adapt https://github.com/ClnViewer/Android-fast-screen-capture/, but I don't really have enough knowledge for that


LmeSzinc:
Rename directory `5` to `Android5` or `Android_5.x-7.x`, and so on.

n0k0m3:
@LmeSzinc done

PR #1240, opened 2022-06-20
Fix nc server loopback IP for AVD (fix #1236)

### LINKED ISSUE #1236: netcat connection refused
### Before you asking

- [X] I have searched the existing issues
- [X] I spend at least 5 minutes for thinking and preparing
- [X] I checked Frequently Asked Questions (FAQ) in Wiki page
- [X] I am using the latest version of Alas

### Describe the bug

Got `nc: connect: Connection refused` when trying to use/benchmark adb_nc on AVD Android 11

### To Reproduce

0. Setup AVD with Android 11, 1280x720
1. Go to benchmark
2. Choose all but hermit
3. Hit "Go"

### Expected behavior

Benchmark successfully run with adb_nc and related tasks

### Relevant log output

```Text
INFO     03:53:27.974 │ BENCHMARK TEST                                                                                          
INFO     03:53:27.975 │ Testing function: screenshot_adb_nc                                                                     
INFO     03:53:27.977 │ Connecting to local emulator, using host 127.0.1.1                                                      
INFO     03:53:27.978 │ Reverse server listening on 127.0.1.1:20154, client can send data to 127.0.1.1:20154                    
WARNING  03:53:32.986 │ b'nc: connect: Connection refused\n'                                                                    
ERROR    03:53:32.989 │ AdbTimeout: reverse server accept timeout                                                               
INFO     03:53:35.994 │ <<< DETECT DEVICE >>>                                                                                   
INFO     03:53:35.996 │ Here are the available devices, copy to Alas.Emulator.Serial to use it or set                           
         Alas.Emulator.Serial="auto"                                                                                            
INFO     03:53:35.999 │ emulator-5554                                                                                           
INFO     03:53:36.001 │ Connecting to local emulator, using host 127.0.1.1                                                      
INFO     03:53:36.003 │ Reverse server listening on 127.0.1.1:20233, client can send data to 127.0.1.1:20233                    
WARNING  03:53:41.013 │ b'nc: connect: Connection refused\n'                                                                    
ERROR    03:53:41.015 │ AdbTimeout: reverse server accept timeout
```

### Screenshots

_No response_

### Anything else?

Using Android Emulator from Android SDK, Android 11 (API 30)

n0k0m3:
This PR attempts to fix #1236 
- Add check for AVD and return loopback IP 10.0.2.2
- Force `host = '127.0.0.1'` for docker setup
  - So far `host = socket.gethostbyname(socket.gethostname())` when used on host + local emulator will return `127.0.0.1` regardless
  - With some linux setup, there is `127.0.1.1 hostname` entry in `/etc/hosts` (for loopback to localhost) which will error out `nc` as above command will return `127.0.1.1`
  - Hardcoding `127.0.0.1` for any emulator setup is sensible

### DIFF
--- a/module/device/connection.py
+++ b/module/device/connection.py
@@ -5,6 +5,7 @@ import socket
 import subprocess
 import time
 import ipaddress
+import platform
 from functools import wraps
 
 import adbutils
@@ -110,6 +111,7 @@ class Connection:
         self.serial = str(self.config.Emulator_Serial)
         self.serial_check()
         self.detect_device()
+        self.is_avd
 
         # Connect
         self.adb_connect(self.serial)
@@ -275,6 +277,16 @@ class Connection:
         return result
 
     @cached_property
+    def is_avd(self):
+        if get_serial_pair(self.serial)[0] is None:
+            return False
+        if 'ranchu' in self.adb_shell(['getprop', 'ro.hardware']):
+            return True
+        if 'goldfish' in self.adb_shell(['getprop', 'ro.hardware.audio.primary']):
+            return True
+        return False
+
+    @cached_property
     def _nc_server_host_port(self):
         """
         Returns:
@@ -290,8 +302,15 @@ class Connection:
         # For emulators, listen on current host
         if self.serial.startswith('emulator-') or self.serial.startswith('127.0.0.1:'):
             host = socket.gethostbyname(socket.gethostname())
+            if platform.system() == 'Linux' and host == '127.0.1.1':
+                host = '127.0.0.1'
             logger.info(f'Connecting to local emulator, using host {host}')
             port = random_port(self.config.FORWARD_PORT_RANGE)
+
+            # For AVD instance
+            if self.is_avd:
+                return host, port, "10.0.2.2", port
+
             return host, port, host, port
         # For local network devices, listen on the host under the same network as target device
         if re.match(r'\d+\.\d+\.\d+\.\d+:\d+', self.serial):


LmeSzinc:
> Hardcoding 127.0.0.1 for any emulator setup is sensible

Nope, from what I have tested (Nox player, LD player, Mumu app player), ip from `socket.gethostbyname(socket.gethostname())` works, `127.0.0.1` doesn't.

n0k0m3:
I guess this is just a Linux issue then.
Given that only AVD, Genymotion, and Anbox are usable on Linux, and they all use 127.0.0.1 instead of hostname ip, I can use `platform.system()` to detect linux, then force `host = '127.0.0.1'` instead.

n0k0m3:
Since the problem is specifically for Linux, I just make a condition flag for the platform and only change the host ip when it reports `127.0.1.1`

LmeSzinc:
Waiting responses from https://github.com/LmeSzinc/AzurLaneAutoScript/issues/1236#issuecomment-1162029223

n0k0m3:
Added a more robust way to detect AVD devices

module/device/connection.py
```diff
@@ -290,8 +291,14 @@ def _nc_server_host_port(self):
...
         if self.serial.startswith('emulator-') or self.serial.startswith('127.0.0.1:'):
             host = socket.gethostbyname(socket.gethostname())
+            if platform.system() == 'Linux' and host == '127.0.1.1':
+                host = '127.0.0.1'
             logger.info(f'Connecting to local emulator, using host {host}')
             port = random_port(self.config.FORWARD_PORT_RANGE)
+            board = self.adb_shell(['getprop', 'ro.hardware.audio.primary'])
+            hardware = self.adb_shell(['getprop', 'ro.hardware'])
+            if 'goldfish' in board or "ranchu" in hardware:
+                return host, port, "10.0.2.2", port
```
LmeSzinc, line 301: This will cost 30ms, and will cost more on slow emulators. Reduce adb_shell calls like
```python
if serial_may_avd(self.serial):
    board = self.adb_shell(['getprop', 'ro.hardware.audio.primary'])
    if 'goldfish' in board:
        return xxx
    hardware = self.adb_shell(['getprop', 'ro.hardware'])
    if "ranchu" in hardware:
        return xxx
return host, port, host, port
```

n0k0m3: AVD uses `127.0.0.1:5555` or `emulator-5554` (so exactly the same as LDPlayer)

n0k0m3:
So set `self.is_avd=False` during `init`, and only check for AVD if `self.is_avd==False` + `emu_device` available and `port_device` is `None`?

LmeSzinc:
```python
@cached_property
def is_avd(self):
    if get_serial_pair()[0] is None:
        return False
    if 'ranchu' in self.adb_shell(['getprop', 'ro.hardware']):
        return True
    if 'goldfish' in self.adb_shell(['getprop', 'ro.hardware.audio.primary']):
        return True
    return False
```

n0k0m3:
Squashed all these commits into 1, should be ready to merge now

PR #1423, opened 2022-07-27
Opt: Modified description of GemsFarming

SaiCateDoan:


module/config/i18n/en-US.json
```diff
@@ -63,7 +63,7 @@
       "help": ""
     },
     "GemsFarming": {
-      "name": "Gems Farming",
+      "name": "Urgent Commission",
```
LmeSzinc, line 66: `Commission Farming`

SaiCateDoan: 修了

PR #1501, opened 2022-08-07
Fix `adb` server version mismatch with docker deployment

n0k0m3:
`adb` shipped with focal-fossa (image base) is too old (version 39), which will cause issue with **all** emulators updated after 2020 (specifically on Linux, using Genymotion and AVD). At the moment only calls made by `adb_command()` is affected as `adb.shell()` actually communicate via socket (affected features: installing `hermittouch` and `ascreencap`)

Use `adb` 1.0.41 from latest `platform-tools` for docker deployment will fix the problem.

I haven't tested this deployment on Windows yet.


LmeSzinc:
> I haven't tested this deployment on Windows yet.

We have auto installer on Windows ~and nobody uses docker on windows~, so this shouldn't be a problem

PR #1560, opened 2022-08-22
Resize WSA window instead of ADB resolution change

Tails32:
I've written a very dirty fix with borrowed code from another python application, so... I guess I'll call it proof-of-concept for now? I'm not familiar with the project or python, but I wanted to test my idea out. It probably introduces an issue as well. If you're interested anyway, here's what I've done:

1. Removed the resizing using ADB commands
2. Added (borrowed) code to resize the window, reposition it and make it borderless
3. Set _screen_size_checked to true in order to let the program run normally from there
4. Added code to trim screenshots to the desired 1280x720 if self.config.Emulator_Serial is "wsa-0"

The result is that ALAS works normally. It recognizes the UI and interacts with it as usual.

I've used this project as the source of the code responsible for resizing the window and make it borderless: https://github.com/banzify/Borderless-Window-Utility
The project does not clearly state any license. Therefore I'm not sure if the code can be used just like this.

### DIFF
--- a/module/device/method/wsa.py
+++ b/module/device/method/wsa.py
@@ -9,6 +9,10 @@ from module.device.method.utils import (RETRY_DELAY, RETRY_TRIES,
 from module.exception import RequestHumanTakeover
 from module.logger import logger
 
+import win32gui
+import ctypes
+import win32con
+import win32api
 
 def retry(func):
     @wraps(func)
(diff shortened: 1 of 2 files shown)


module/device/method/wsa.py
```diff
@@ -9,6 +9,10 @@
 from module.exception import RequestHumanTakeover
 from module.logger import logger
 
+import win32gui
+import ctypes
+import win32con
+import win32api
```
LmeSzinc, line 15: Import these at the beginning of `display_resize_wsa()`, or linux will import the dependencies from windows

module/device/method/wsa.py
```diff
@@ -145,5 +149,30 @@ def get_display_id(self):
...
+            user32 = ctypes.windll.user32
+            user32.SetWindowLongW(hwnd, win32con.GWL_STYLE, win32con.WS_VISIBLE | win32con.WS_CLIPCHILDREN)
+            user32.SetWindowLongW(hwnd, win32con.GWL_EXSTYLE, 0)
+            HRes = int(1280)-1
+            VRes = int(720)-1
+            user32.MoveWindow(hwnd, XPos, YPos, HRes, VRes, True)
+            VRes = VRes + 1
+            HRes = HRes + 1
+            win32gui.SetWindowPos(hwnd, None, XPos, YPos, HRes, VRes, win32con.SWP_FRAMECHANGED | win32con.SWP_NOZORDER |
+                                  win32con.SWP_NOOWNERZORDER)
```
LmeSzinc, line 178: 1. Code need to be pythonic
2. DPI scaling is not calculated when resizing the window, thanks @hgjazhgj

module/device/screenshot.py
```diff
@@ -93,6 +94,28 @@ def _handle_orientated_image(self, image):
...
+
+        if not self._screen_size_checked:
+            return image
+
+        if self.config.Emulator_Serial != 'wsa-0' or (width == 1280 and height == 720):
+            return image
+
+        # Trim screenshots only after screen size was checked
+        # and if they are both from WSA and not already 1280x720
+        image = image[0:720, 0:1280]
```
LmeSzinc, line 115: Use the `crop()` function from module.base.utils

module/device/method/wsa.py
```diff
@@ -145,5 +149,30 @@ def get_display_id(self):
...
+        def winEnumHandler(hwnd, ctx):
+            if win32gui.IsWindowVisible(hwnd):
+                n = win32gui.GetWindowText(hwnd)
+                s = win32api.GetWindowLong(hwnd, win32con.GWL_STYLE)
+                x = win32api.GetWindowLong(hwnd, win32con.GWL_EXSTYLE)
+                if n:
+                    allvisiblewindows.update({n: {'hwnd': hwnd, 'STYLE': s, 'EXSTYLE': x}})
+        allvisiblewindows = {}
+        win32gui.EnumWindows(winEnumHandler, None)
+        hwnd = allvisiblewindows.get('Azur Lane',{}).get('hwnd')
```
LmeSzinc, line 161: Multi-server support required, search `碧蓝航线`, `アズールレーン`,  `碧藍航線` also

LmeSzinc (commented):
Still need more feedback from WSA users to merge this PR.

module/device/method/wsa.py
hgjazhgj: also, can this try-GetWindowRect be replaced with if-IsWindow?

mxdongfang:
> 仍然需要 WSA 用户的更多反馈来合并此 PR。

wsa 2308原版使用此修复能正常运行，暂时没有遇到bug，就是每次启动程序都需要重新修改文件
image

PR #1923, opened 2022-11-30
Add: Support OS Voucher Shop

nEEtdo0d:
- Implements module to support the os voucher shop interface, near identical copy to medal shop
- Other servers will need to capture their own `EXCHANGE_ENTER` ~~and `SHOP_BUY_CONFIRM_VOUCHER`~~. Other assets should be server transparent but feel free to update.
- Add supporting OS tasks OpsiVoucher, configured as monthly after a successful run
- OpsiVoucher is for purchasing all items except for logger archives and also special radar since that uses a different cost

Unused but code retained in case becomes a trend. Currently is a one-time feature that AL devs impose, does not refresh monthly
- Add supporting OS funcs OpsiArchive, configured as monthly after a successful run (will need additional edits for complete insertion)
- Would be split as a secondary task due to a restriction AL allowing purchase of 1 logger archive type at a time
- OpsiArchive completes the active logger archive, purchases available logger archive, and repeats until none left

Will update wiki on shop information to include list of voucher shop items


module/shop/shop_voucher.py
```diff
@@ -0,0 +1,230 @@
...
+        if self.appear(SHOP_BUY_CONFIRM_SELECT, offset=(20, 20), interval=3):
+            self.shop_buy_select_execute(item)
+            self.interval_reset(SHOP_BUY_CONFIRM_SELECT)
+            return True
+        if self.appear(SHOP_BUY_CONFIRM_AMOUNT, offset=(20, 20), interval=3):
+            self.shop_buy_amount_execute(item)
+            self.interval_reset(SHOP_BUY_CONFIRM_AMOUNT)
+            return True
+        if self.appear_then_click(SHOP_BUY_CONFIRM_VOUCHER, offset=(20, 20), interval=3):
+            return True
```
LmeSzinc, line 170: Try to reuse `handle_popup_confirm()` as `SHOP_BUY_CONFIRM_VOUCHER`

module/shop/shop_voucher.py
```diff
@@ -0,0 +1,230 @@
...
+        if not self.shop_filter:
+            return
+
+        # When called, expected to be in
+        # correct Voucher Shop interface
+        logger.hr('Voucher Shop', level=1)
+        self.wait_until_voucher_appear()
+
+        # Execute buy operations
+        while 1:
```
LmeSzinc, line 188: Add before loop:
```python
MATERIAL_SCROLL.set_top(main=self)
```

module/shop/shop_voucher.py
nEEtdo0d: Did while debugging but just now realized I could temporarily update `_popup_offset` if that is acceptable?

I see in `retirement` does something similar `backup, self._popup_offset = self._popup_offset, (20, 50)`

module/shop/shop_voucher.py
nEEtdo0d: Can do, guess I can update `MedalShop` with the same since I took it from there. So if it didn't have it here, then likely not there either

module/shop/shop_voucher.py
LmeSzinc: Add argument `offset` to `handle_popup_*` would be nice, see `handle_popup_single`. Modify dirty patches on `_popup_offset` too

nEEtdo0d: Done, updated func signatures and fixed dirty patches only `retirement` did this
Removed the unneeded asset

module/shop/shop_voucher.py
nEEtdo0d: Done, added to both

PR #2165, opened 2023-01-23
Add: Raid EX Support

nEEtdo0d:
Since dev is behind and my development needed recent commits for testing, branched from master. Can re-direct to bug_fix if that is clear to merge master into it.

With a beefy enough team plus submarine, EX can be cleared somewhat consistently. Generally getting old to do this manually nowadays.

- Support automate EX mode (working and tested draft)
    - Focus was re-use where possible but can re-factor to a different approach
    - Changed Combat inheritance to GuildCombat which is backwards compatible.  As EX doesn't use standard assets (not S, A, B ranking)
- EN shipyard asset updates, position changed this affects the OCR read as a result
- GuildCombat variant, thought it was strange one of the funcs did not call its super method
    - This also helps with the inheritance change with EX using its specific combat end assets while being backwards compatible to its standard assets


module/raid/raid.py
```diff
@@ -246,6 +249,13 @@ def raid_execute_once(self, mode, raid):
             Campaign_UseAutoSearch=False,
             Fleet_FleetOrder='fleet1_all_fleet2_standby'
         )
+
+        if mode == 'ex':
+            self.config.override(
+                Submarine_Fleet=1,
+                Submarine_Mode='every_combat'
+            )
```
LmeSzinc, line 257: 1) `self.config.override` will last until task end, may cause unexpected behaviour.
2) `Submarine_Fleet` and `Submarine_Mode` can still be set to const value without adding argument group `Submarine` to task `Raid`

module/raid/run.py
```diff
@@ -35,6 +73,12 @@ def run(self, name='', mode='', total=0):
         if not name or not mode:
             raise ScriptError(f'RaidRun arguments unfilled. name={name}, mode={mode}')
 
+        if mode == 'ex':
+            self.ui_goto_main()
+            Reward(self.config, self.device).reward_mission(
+                daily=self.config.Reward_CollectMission,
+                weekly=False)
```
LmeSzinc, line 80: Too many `Reward` runs. Move to `RaidDaily`, check once before running EX dailies

module/raid/raid.py
nEEtdo0d: 1. That's right usually should have put in a sort of before and after restore back to its original value I thought `override` did this. Is this alright?
```
backup = self.config.Submarine_Fleet
self.config.Submarine_Fleet = 1
...
self.config.Submarine_Fleet = backup
```
2. Adding to the argument group was for users that needed submarine to be configurable since `Raid`. I will remove the argument group since I guess in general you'd want to avoid using submarine in other difficulties besides EX mode.

module/raid/run.py
nEEtdo0d: I had inserted into here so that it was applicable to both tasks `Raid and RaidDaily`. But I think you are right, if user is running `Raid` themselves then its assumed they collected have the tickets themselves, no need to have alas retrieve them.  But shouldn't be after `RaidDaily` since tickets can only be acquired after doing 10+ runs

Another idea I had  a new task `RaidDailyEX`. Order would be `RaidDaily > self.config.task_call('Reward') > RaidDailyEX`. A little longer I think it would take but more naturally calls `Reward` this way.

Can't call `task_call` and have steps after it since raises an exception. I believe but if separated as two tasks it would work I think.

module/raid/raid.py
```diff
@@ -126,7 +129,7 @@ def pt_ocr(raid):
         return None
 
 
-class Raid(MapOperation, Combat, CampaignEvent):
+class Raid(MapOperation, GuildCombat, CampaignEvent):
```
nEEtdo0d, line 132: Did this for quickness but was wondering should make a `RaidCombat` equivalent? Or should update `Combat` with the extra clauses since it is backwards compatible but maybe the assets are too similar to others?

module/raid/raid.py
LmeSzinc: 1.
```python
backup = self.config.temporary(key=value)
...
backup.recover()
```
2. Normal battles have submarine hunt and calls, but raid can dont have submarine hunt. If users have their submarine fleet prepared in raid, it's prefered to call them in combat, otherwise users just clear out the submarine fleet. So I think it unnecessary to have a standalone config.

module/raid/raid.py
LmeSzinc: Make a `RaidCombat`

nEEtdo0d:
Thanks for the review, think it is good now.

PR #2173, opened 2023-01-26
Opt: Add optional MAA installation for Docker

kmou424:
Try to add MAA support when building alas docker image

Users should download and extract MAA latest release archive to a specific path(`../MAA`), and then uncomment `# - '../MAA:/app/MAA'` in `docker-compose.yml`, and now they can build a new alas image with MAA support.
In addition, they need to change the `MaaEmulator.MaaPath.name` to `/app/MAA` in MAA configuration.

Tip. Since the MAA latest release is built by g++12 and requires GLIBC 2.32 or higher, we changed the base docker image to ubuntu jammy(22.04 LTS), which already have prebuilt GLIBC 2.35

### DIFF
--- a/deploy/docker/Docker-run.sh
+++ b/deploy/docker/Docker-run.sh
@@ -51,4 +51,8 @@ prun_or_continue "adb kill-server"
 
 pprint "Running the container"
 trap "rm ${XDG_RUNTIME_DIR}/${CONTAINER}.lock && docker kill ${CONTAINER}" EXIT
+
 prun "docker run --net=host --volume=${SOURCE}/..:/app/AzurLaneAutoScript:rw --interactive --tty --name ${CONTAINER} ${CONTAINER}"
+# If you need MAA support, uncomment the following two lines and comment the line above(Modify the path of MAA according to the actual situation)
+# MAA_SOURCE="${SOURCE}/../../MAA"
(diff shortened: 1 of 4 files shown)


LmeSzinc:
Create new docker file for MAA and keep `docker-run.sh`, since I dont know how many people is using the olds.

kmou424:
> Create new docker file for MAA and keep `docker-run.sh`, since I dont know how many people is using the olds.

I've restored the old `Docker-run.sh` as you suggested.
The installation of MAA will not be affected by the `Dockerfile`, so we don't need to create a new `Dockerfile` for MAA separately. My only change to the `Dockerfile` is to replace the old base image with a newer base image, in order to be able to run MAA.
If users want optional MAA installation, just uncomment the line containing `volume="*MAA"` in `docker-compse.yml` (or `Docker-run.sh`).

LmeSzinc:
Alas requires `git`, it was in `mambaforge` image but haven't been installed in the new dockerfile

kmou424:
> Alas requires `git`, it was in `mambaforge` image but haven't been installed in the new dockerfile

Yes, this is my mistake. Do I need to open a new pull request to modify the dependencies installation?

kmou424:
> Alas requires `git`, it was in `mambaforge` image but haven't been installed in the new dockerfile

Cherry-pick it plz: https://github.com/kmou424/AzurLaneAutoScript/commit/76185796d6a1ed4b9f1788284a4ca5a5f472c582

LmeSzinc:
Create a new PR adding `git`

PR #2225, opened 2023-02-09
Opt: Extremely reduce image size built by Dockerfile (-59%)

hgjazhgj:
This is the Dockerfile I have used in Docker Desktop with Hyper-V backend, Win 10.  
See [https://github.com/hgjazhgj/Alas_Docker_BlueStacks5](https://github.com/hgjazhgj/Alas_Docker_BlueStacks5).  
image
The image size is reduced from 2.44GB to 1GB.  
I am not sure how to do for Dockerfile.cn, it seems that the origin Dockerfile will download some files from GitHub..., and apt sources are not changed.  
I also have Dockerfile for AArch64, running in my RaspberryPi, but using 3rd party MXNet from [https://www.binss.me/blog/run-azurlaneautoscript-on-arm64/](https://www.binss.me/blog/run-azurlaneautoscript-on-arm64/), dose it have the chance to merge into Alas?

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -1,46 +1,18 @@
-FROM ubuntu:jammy
+# docker build -t hgjazhgj/alas:latest .
+# docker run -v ${PWD}:/app/AzurLaneAutoScript -p 22267:22267 --name alas -it --rm hgjazhgj/alas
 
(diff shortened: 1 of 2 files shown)


hgjazhgj:
Done

PR #2297, opened 2023-02-27
Fix: sync python executable with python 3.7 slim image

kmou424:
Update the python executable in the template after the docker image of python changed


LmeSzinc:
配置文件模板是自动生成的，需要改一下生成的代码
https://github.com/LmeSzinc/AzurLaneAutoScript/blob/e78d83bc0a62339496d9b41553d74c06a1d100d4/module/config/config_updater.py#L401-L433

PR #2327, opened 2023-03-03
Add: Disable flagship and vanguard change in GemsFarming

Alexan-ded:
Added "Disable" options for low emotion farming.
I don't know its Chinese translation, so kept it default for now. Some "help" text changes may be needed.
Also I changed module/config/redirect_utils/utils.py accordingly, but I have no idea if it works.
And I brought Enemy Priority menu back, sometimes I farm difficult stages (for 0 LB frontline) and use < 33 level backline, so killing only 1-2 star enemies keeps my fleet alive.


Alexan-ded:
> To keep the settings simple, remove `EnemyScaleBalanceWeight` and override it to `S1_enemy_first` in code.

How does having `EnemyScaleBalanceWeight` increase the difficulty of the settings? Every other category in "Farm" has it, having it here provides overall more control over farming, so I see no reason to remove it.

Also what should I write as the Chinese translation for `Disabled`? And is my utils.py change any good?

LmeSzinc:
GemsFarming is for 0LB farming, easy enemies are prefered. It's redundant to have a `EnemyScaleBalanceWeight` setting in GemsFarming letting users to choose since they will all choose `S1_enemy_first`. Hardcoding the setting reduces usage costs.

> Also what should I write as the Chinese translation for Disabled? Is 禁用 ok?

I prefer `Don't Change` in English and `不更换` in Chinese like the existing options.

module/config/redirect_utils/utils.py
```diff
@@ -86,7 +86,9 @@ def change_ship_redirect(value):
       FlagshipChange + FlagshipEquipChange -> ChangeFlagship
     """
     ship, equip = value
-    if equip:
+    if ship == 'disabled':
+        return 'disabled'
+    elif equip:
         return 'ship_equip'
     else:
         return 'ship'
```
LmeSzinc, line 94: Variable `ship` and `equip` are from old configs, both of them are `bool`. So:
```python
if not ship:
    return 'disabled'
elif ...
```

PR #2395, opened 2023-03-24
Fix weird 755 mode on png's in asset/jp/campaign

hgjazhgj:
These 4 images in `asset/jp/campaign` have weird modes of 755 other than any other png's in alas of 644
```
 mode change 100755 => 100644 assets/jp/campaign/SWITCH_1_HARD.png
 mode change 100755 => 100644 assets/jp/campaign/SWITCH_1_NORMAL.png
 mode change 100755 => 100644 assets/jp/campaign/SWITCH_2_EX.png
 mode change 100755 => 100644 assets/jp/campaign/SWITCH_2_HARD.png
```
Why does an IMAGE need EXECUTE permission?
When I think about it, I feel uncomfortable all over, as if ants are crawling on my body
SO FIX IT


SaiCateDoan:
F WORD

PR #2428, opened 2023-03-30
Fix: MEOWFFICER_FEED_CANCEL click fails

Alexan-ded:
MEOWFFICER_FEED_CANCEL click takes a MEOWFFICER_ENHANCE_CONFIRM as a check_button, but (at least in EN) MEOWFFICER_ENHANCE_CONFIRM is MEOWFFICER_FEED_CONFIRM shifted by ~20 px, so the click fails


guoh064:
Remembered similar issues in JP, after checking enhance materials alas sometimes do not confirm but cancel instead.

PR #2551, opened 2023-05-03
Fix: Redundant `template-maa.maa.json` config file

hgjazhgj:
I have noticed that when I add a new MAA config file in alas webui, a redundant `template-maa.maa.json` config file will be created and displayed as `template-maa`.
It takes me a long time to debug... Finally I find that it is the fault of incorrect `is_template_config` judgment when there the template filename is one of `template` `template-alas` `template-maa`.

The files of `template-alas` and `template-maa` does not actually exist, but they will be passed to `*Config.__init__` by `AlasGUI.ui_add_alas`. and then, `AzurLaneConfig.is_template_config` is False and `AzurLaneConfig.save()` will be called.

`AzurLaneConfig.save()` would not write the disk when there is no modification. But in the situation of `template-maa`, `AzurLaneConfig.modified` is as below:

```json
{
    "Commission.Scheduler.Enable": true,
    "Research.Scheduler.Enable": true,
    "Reward.Scheduler.Enable": true
}
```

so a `template-maa.maa.json` config file will be created.


hgjazhgj:
我其实一直以为新建配置的时候传的template-alas和template-maa会有什么魔法转换成对应的文件名,想不到是嗯传(

PR #2872, opened 2023-07-13
Add: Blueprint assets of research S6

bookbug666:


assets/research_blueprint/felix.png
LmeSzinc, line 1: schultz.png

bookbug666: done

PR #2978, opened 2023-08-05
Upd: [EN] DISASSEMBLE_CANCEL

Alexan-ded:


LmeSzinc:
`DISASSEMBLE_CONFIRM` and `DISASSEMBLE_CANCEL` are a pair, check if `DISASSEMBLE_CONFIRM` need update

Alexan-ded:
> `DISASSEMBLE_CONFIRM` and `DISASSEMBLE_CANCEL` are a pair, check if `DISASSEMBLE_CONFIRM` need update

It is slightly offset from the previous version, but it does work. Updated anyway

PR #3381, opened 2023-12-26
Upd: update WSA support base on PR#1560

q609464692:
Updated WSA support features base on [PR#1560](https://github.com/LmeSzinc/AzurLaneAutoScript/pull/1560)


module/device/method/wsa.py
```diff
@@ -145,5 +145,35 @@ def get_display_id(self):
...
+                if n:
+                    allvisiblewindows.update({n: {'hwnd': hwnd, 'STYLE': s, 'EXSTYLE': x}})
+        allvisiblewindows = {}
+        hwnd = 0
+        multi_server_name = ['Azur Lane', '碧蓝航线', 'アズールレーン', '碧藍航線']
+        win32gui.EnumWindows(winEnumHandler, None)
+        for i in multi_server_name:
+            if i in allvisiblewindows:
+                hwnd = allvisiblewindows.get(i).get('hwnd')
+                break
```
hgjazhgj, line 166: I would like to ask why not use `FindWindow` here?

hgjazhgj:
You need to UNDERSTAND the code and edit it by yourself but not just COPY them

module/device/method/wsa.py
q609464692: sure, updated

hgjazhgj (changes requested):
I have said *You need to UNDERSTAND the code and edit it by yourself but not just COPY them* before.
You have never understand these code. Neither responsing to previous questions in #1560 (e.g. calc high DPI scaling), nor writing responsible code (e.g. win32gui hasattr SetWindowLong).
Such behavior is ill-intentioned. Never message me again!
Blocked.

q609464692:
Good scolding！
Perhaps I should request a merger with the previous PR owner first so as not to cause such great anger. I just want to complete the PR that the previous author did not continue with, so that ALAS can support WSA, but I haven't spent more time delving deeper into it.
I'll try to change the code according to the requirements of the final reviewer, but I'll not open this PR again. Thanks for the review.

PR #3420, opened 2024-01-10
feat: add simple auto start emulator

not-rzwei:
## Changes
- Add new setting to auto start the emulator when we click the Start button
- Translation for English and Japanese for above feature

image

## Why
I saw there is on-hold initiative for restarting emulator but I found it too complicated when the user can supply the command by themselves. The idea is from MAA with their startup emulator feature, though it might not be ideal for use case like cloud phone.


module/device/emulator.py
```diff
@@ -323,3 +327,18 @@ def emulator_restart(self):
...
+            return
+
+        logger.info('Starting emulator')
+        try:
+            self.execute(command)
+        except Exception:
+            logger.error(f'Cannot execute start command: {command}')
+
+        if isinstance(self.config.StartEmulator_Delay, float) and self.detect_emulator_status(self.serial) == 'offline':
+            time.sleep(self.config.StartEmulator_Delay)
```
LmeSzinc, line 344: Using sleep is forbidden.
- Check if emulator has completed started up (ready for shell commands)
- Retry if emulator hang or died during startup

not-rzwei: I've updated to wait until the the emulator is connected to ADB since it's more or less the same as checking the shell command. https://github.com/LmeSzinc/AzurLaneAutoScript/pull/3420/commits/be4a8843dacb91e548c16aaead8f096b2d859374#diff-0c44e99053d6d86441db43976c180bbc1c932e5fc7a476c34b64ec6942614f0cR350

As for retry emulator, I was thinking of leaving that part to user, if it's hanging or died. The user should check themselves either there's something wrong with the emulator, the command, or it's just taking too long to boot. I wanted to make it simple for now but let me know if you have different thought on this.

PR #3817, opened 2024-05-26
Update alas.py

Condor2436:
add critical log recording


Condor2436:
head branch selected by mistake

PR #3974, opened 2024-07-06
fix: opsi amount handler and counter ocr

SaarChaffee:
- amount handler
- counter ocr should return number
- counter area

### DIFF
--- a/module/os_shop/item.py
+++ b/module/os_shop/item.py
@@ -43,10 +43,10 @@ class CounterOcr(Ocr):
         result_list = super().ocr(image, direct_ocr=direct_ocr)
         if isinstance(result_list, list):
             for r in range(len(result_list)):
-                result_list[r] = result_list[r].split('/')
+                result_list[r] = [int(i) for i in result_list[r].split('/')]
             return result_list
         else:
-            return result_list.split('/')
+            return [int(i) for i in result_list.split('/')]
 
 
 COUNTER_OCR = CounterOcr([], threshold=96, name='Counter_ocr')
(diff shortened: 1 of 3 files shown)


module/os_shop/item.py
```diff
@@ -43,10 +43,10 @@ def ocr(self, image, direct_ocr=False):
         result_list = super().ocr(image, direct_ocr=direct_ocr)
         if isinstance(result_list, list):
             for r in range(len(result_list)):
-                result_list[r] = result_list[r].split('/')
+                result_list[r] = [int(i) for i in result_list[r].split('/')]
             return result_list
```
hgjazhgj, line 47: return [[int(j)for j in i.split('/')]for i in result_list]

module/os_shop/shop.py
```diff
@@ -96,29 +97,57 @@ def shop_buy_amount_handler(self, item):
             ScriptError: OCR_SHOP_AMOUNT
         """
         currency = self.get_currency_coins(item)
+        count = min(int(currency // item.price), item.count)
```
hgjazhgj, line 100: currency和item.price不应该是浮点数吧
int间的__truediv__返回的就是int

module/os_shop/item.py
SaarChaffee: lme说不缺这点行数

module/os_shop/shop.py
SaarChaffee: 我也不知道为什么要加个 int https://github.com/LmeSzinc/AzurLaneAutoScript/blob/6ac38928bd713fb74abb391b113a0e72a61cdd68/module/shop/clerk.py#L246

module/os_shop/item.py
LmeSzinc: 屑，快改，python下
```
for i in range():
    list[i] ...
```
要打屁股的

SaarChaffee: 诶

PR #4069, opened 2024-08-08
Fix: Minigame bugfix after content expansion

yypsybs:
修复小游戏增加导致的bug


1533258591:
[log.txt](https://github.com/user-attachments/files/16591211/log.txt)

PR #4433, opened 2024-12-14
Fix: to_map_input_name case issue

Shane-Xue:
After bugfixes #4423 logic in the function changed and on the line where CAMPAIGN_ is removed the name is actually in lower case. 
Moved upper case manupulation in order to ensure correct case.  
Upper was not removed considering the function should turn d3 into D3. Before this bugfix incrementing to the next stage was unusable (eg. 12-2 would be incremented to a strange "n2")


Shane-Xue:
I pulled this branch from the master branch but I think I should PR it into dev, I'm not sure if this is allowed or if I should redo anything.

Issue #4463, opened 2024-12-25
High-Efficiency Combat Logistics plan bug EN server


dragonheart107:
### Before you asking

- [X] I have searched the existing issues
- [X] I spend at least 5 minutes for thinking and preparing
- [X] I checked Frequently Asked Questions (FAQ) in Wiki page
- [X] I am using the latest version of Alas

### Describe the bug

example for bug is current event d3, however it happens in campaign stages as well:
After having finished a High-Efficiency Combat Logistics plan (referred to as "2x loot" going forward) once; on the post stage pop-up where you can repeat the stage:
![MuMu12-20241225-103309](https://github.com/user-attachments/assets/34994487-58db-4d49-89b6-066ef4267a43)
ALAS correctly ticks the 2x loot option and stage starts as normal; however while Azur Lane is correctly farming the stage, ALAS prints out below log after a few seconds of starting the stage and wants to force restart even though Azur Lane is running correctly,

### To Reproduce

1. Select any stage where you can use 2x loot and have the nonstop option in alas or at least run in more than once -> let it finish once
2. ALAS ticks 2x loot
3. Azur Lane goes into the stage and starts auto search
4. ALAS prints below log after a few seconds

### Relevant log output

```Text
═══════════════════════════════════════ D3 ════════════════════════════════════════
INFO     16:28:14.410 │ D3                                                         
INFO     16:28:14.410 │ Count: 1                                                   
INFO     16:28:14.411 │ In auto search menu, skip ensure_campaign_ui.              
─────────────────────────────────────── D3 ────────────────────────────────────────
INFO     16:28:14.415 │ D3                                                         
INFO     16:28:14.415 │ Expect emotion reduce: (24, 4)                             
INFO     16:28:14.422 │ Bind task ['General', 'Alas', 'TaskBalancer',              
         'EventGeneral', 'Event']                                                  
INFO     16:28:14.423 │ Save config ./config\alas.json,                            
         Event.Emotion.Fleet1Value=130,                                            
         Event.Emotion.Fleet1Record=datetime.datetime(2024, 12, 25, 16, 28, 14),   
         Event.Emotion.Fleet2Value=115,                                            
         Event.Emotion.Fleet2Record=datetime.datetime(2024, 12, 25, 16, 28, 14)    
INFO     16:28:14.427 │ [Emotion fleet_1] 130                                      
INFO     16:28:14.428 │ [Emotion fleet_2] 115                                      
[...]

PR #4478, opened 2024-12-30
remove outdated EN -Mail assets,  enable EN AUTO-Dossier

dragonheart107:
Please double-check just in case, but should be fine.
(Ignore the few commits of deleting and adding, was testing stuff and saving in wrong places, couldn't yet test auto dossier though).


dragonheart107:
it seems that auto dossier struggles as it doesn't realize that it's already in progress

dragonheart107:
i think the new asset should fix the issue but need to test tomorrow

campaign/campaign_main/campaign_14_4.py
```diff
@@ -133,19 +123,13 @@ def battle_3(self):
...
-        if self.clear_filter_enemy(self.ENEMY_FILTER, preserve=0):
-            return True
+        self.pick_up_ammo(H5)
 
         return self.battle_default()
 
     def battle_7(self):
-        self.fleet_boss.pick_up_flare(A5)
 
         return self.fleet_boss.clear_boss()
```
LmeSzinc, line 134: The original battle functions are better.
- Picking ammo and flare at battle_3 (after the 4th battle) would avoid empty ammo debuff (-50% dmg) in the next 3 battles and gain ammo buff (+10% dmg)
- `clear_filter_enemy` preserves the weakest enemy and release it at the last battle
- Having flares on boss fleet makes it faster to clear mob waves before the real boss and less vanguard HP loss

LmeSzinc:
When having multiple standalone changes, each of them should have a separated PR

dragonheart107:
> When having multiple standalone changes, each of them should have a separated PR

Okay, I'll do that in the future

campaign/campaign_main/campaign_14_4.py
dragonheart107: The issue with the old code was however that the fleet blocking the lighthouse would often not be engaged first, missing out a lot on the debuff.
For the flares being split between mob and boss it kinda depends on your fleets.

dragonheart107: > The original battle functions are better.
> - Picking ammo and flare at battle_3 (after the 4th battle) would avoid empty ammo debuff (-50% dmg) in the next 3 battles and gain ammo buff (+10% dmg)

That was something mine did too, though the order when there is no dmg buff is rearranged

Old code:
5ammo -> Battle_0 dmg buff -> 4ammo
4ammo -> Battle_1 dmg buff -> 3ammo
3ammo -> Battle_2 **no dmg buff** -> 2ammo
2ammo -> Battle_3 ammo refill -> 5ammo -> dmg buff -> 4ammo
4ammo -> Battle_4 dmg buff -> 3ammo
3ammo -> Battle_5 dmg buff -> 2ammo
2ammo -> Battle_6 **no dmg buff** -> 1ammo

Mine:
5ammo ->  B0 dmg buff -> 4ammo
4ammo -> B1 dmg buff -> 3ammo
3ammo ->  B2 ammo refill -> 5ammo -> dmg buff -> 4ammo
4ammo ->  B3 ammo refill -> 5ammo ->dmg buff -> 4ammo
4ammo ->  B4 dmg buff ->3ammo
3ammo ->  B5 **no dmg buff** ->2ammo
2ammo ->  B6 **no dmg buff** -> 1ammo

Probably ends up being an issue of preference

dragonheart107:
not sure what the git etiquette is, do you want me to close this pull request or keep open? @LmeSzinc

dragonheart107:
Okay so I have a few further questions, because it seems I misunderstood a few things regarding pick_up_ammo:

During event downtime I was able to run both the original `campaign_14_4.py` and a further modified one:
```

    def battle_0(self):
        self.pick_up_light_house(A9)
        self.pick_up_flare(A5)

        if self.clear_roadblocks([road_A8, road_H9], weakest=False):
            return True

        return self.battle_default()

    def battle_1(self):
        self.pick_up_light_house(A9)
        
        if self.clear_filter_enemy(self.ENEMY_FILTER, preserve=1):
            return True

        return self.battle_default()

    def battle_2(self):
        self.pick_up_ammo(H5)

        if self.clear_filter_enemy(self.ENEMY_FILTER, preserve=1):
            return True

        return self.battle_default()
    
    def battle_3(self):
        self.pick_up_ammo(H5)

        if self.clear_filter_enemy(self.ENEMY_FILTER, preserve=1):
            return True

        return self.battle_default()
    #battle_3 and battle_4 probably redundant

    def battle_4(self):

        if self.clear_filter_enemy(self.ENEMY_FILTER, preserve=0):
            return True

        return self.battle_default()

    def battle_7(self):
        self.pick_up_flare(H9)

        return self.fleet_boss.clear_boss()
    
```

And during both I noticed that there are issues with pick_up_ammo:
It doesn't allow for 'layered' pick up as it will only run pick up ammo once and not a second time even if there is still 1 ammo there;  even if you place a coordinate inside (like `self.pick_up_ammo(H5)`) it is ignored. Is this due to the `recover`-definition in `map.py` (see below)?
It also doesn't show as MA in the log output global map anymore.
It sometimes doesn't pick up any ammo at all, maybe issue with 14_4 specific `OVERRIDE` ?
I also ran 12_4 once to further understand the ammo behaviour and there it picked up ammo at `battle_3`  but not again even if it should go there again at `battle_6` and there was 1 ammo left to pick-up.


From map.py
```

    def pick_up_ammo(self, grid=None):
        """
        Args:
            grid (GridInfo):
        """

        if grid is None:
            grid = self.map.select(may_ammo=True)
            if not grid:
                logger.info('Map has no ammo.')
                return False
            grid = grid[0]

        if self.ammo_count > 0 and grid.is_accessible:
            logger.info('Pick up ammo: %s' % grid)
            self.goto(grid, expected='')
            self.ensure_no_info_bar()

            # self.ammo_count -= 5 - self.battle_count
            recover = 5 - self.fleet_ammo
            recover = 3 if recover > 3 else recover
            logger.attr('Got ammo', recover)

            self.ammo_count -= recover
            self.fleet_ammo += recover
```
I'll probably go get some more testing done when there isn't a rerun or event ongoing. Would be glad to hear your thoughts/ explanation.

LmeSzinc:
Mail and dossier updates are ready to merge, 14-4 needs further discussion.
When having multiple changes in one PR this is what would happen: one change blocks all changes from be merged.

dragonheart107:
> Mail and dossier updates are ready to merge, 14-4 needs further discussion. When having multiple changes in one PR this is what would happen: one change blocks all changes from be merged.

for now i'm keeping the 14-4 changes locally, until i have more testing data

dragonheart107:
After I had a look through the different assets due to #4463, I was wondering how exactly ALAS knows if it's supposed to use `mail.py` or `mail_white.py` . As far as I can tell the white mail was originally a fix for JP, but according to the error logs it is also called in en server and since there the asset `MAIL_WHITE_EMPTY.png` is missing it gives an error. Though this happens only sometimes.

PR #4590, opened 2025-02-15
Help needed for a feature for augment conversion

dragonheart107:
I am kinda stuck at this point and I think I simply lack the knowledge of fixing it (also sorry in advance, for the not cleaned up code and comments yet):

1. Functional Issue with the current code (refer to module/augment/conversion_unique.py):

line 127 onwards: it often times just goes to self.augment_popup_close(), seeming to not recognize the previous ifs in the while loop

But when it does recognize them correctly, it results in failing to properly leave the popup after and instead just goes directly to swiping
see this recording:

https://github.com/user-attachments/assets/087562e1-3512-4ea4-9ee2-d5d637e1d589

2. In module/augment/conversion_unique.py line 215 onwards:

```
    def conv_uniq_cycle(self, skip_first_screenshot=False):
        logger.hr('Augment conversion: Uniques', level = 3)
        interval =Timer(3)
        while interval.reached():
            if skip_first_screenshot:
                skip_first_screenshot = False
            else:
                self.device.screenshot()

            if self.is_in_gear():
                self.conv_uniq_enhance()
                continue
            if self._ship_view_swipe(-400) is True:
                continue
            if self._ship_view_swipe(-400) is False:
                break
                
```
How should i code it so that it stays in conv_uniq_enhance cycle as long as we are in gear and if the cycle is done it swipes once.
then if it is a different ship start again from conv_uniq_enhance;
however if ship stays the same after swipes break out of loop.

I appreciate every input for this.


module/augment/conversion_unique.py
```diff
@@ -0,0 +1,321 @@
...
+            if skip_first_screenshot:
+                skip_first_screenshot = False
+            else:
+                self.device.screenshot()
+            self.not_appear_then_click(AUGMENT_EMPTY, AUGMENT_INGEAR_CONFIRM, interval=3)
+
+            if skip_first_screenshot:
+                 skip_first_screenshot = False
+            else:
+                 self.device.screenshot()
```
LmeSzinc, line 114: 1. Your code is filled with click-sleep thingy, operation success is not guaranteed
see https://github.com/LmeSzinc/AzurLaneAutoScript/wiki/4.4.-State-loop

module/augment/conversion_unique.py
```diff
@@ -0,0 +1,321 @@
...
+                if self.handle_convt2_finish():
+                    logger.hr('Conversion in progress')
+                    continue
+                #actual conversion enhancing part
+                if self.appear(CONVT2_START, interval=3):
+                    logger.hr('Conversion start')
+                    self.appear_then_click(CONVT2_START_COST, interval=3)
+                 #interval.reset()
+                    self.appear_then_click(CONVT2_START, interval=3)
+                    self.appear_then_click(CONVT2_NEW, interval=3)
```
LmeSzinc, line 144: same as above

LmeSzinc:
2. Exit to dock and re-enter first ship after argument function calls, as game client would guarantee next ship to be the first. It's hard to detect swiping to a *different* ship, because users can have duplicate ships having same UI appearance.
Check the implementation of awakening: https://github.com/LmeSzinc/AzurLaneAutoScript/blob/aa12f86ea91764909a8d82b4bc6ae3a6e8dcbc44/module/awaken/awaken.py#L340-L364

module/augment/conversion_unique.py
dragonheart107: should these then be something like this part from the 4.4 State loop doc? or does that structure not change much

```
       if self.appear(DORM_MANAGE_CHECK, offset=(20, 20)):
            break
        if self.appear_then_click(DORM_MANAGE, offset=(20, 20), interval=2):
            continue
        if self.ui_additional():
            continue
        if self.appear_then_click(DORM_FURNITURE_CONFIRM, offset=(30, 30), interval=2):
            continue
```

dragonheart107:
> 2. Exit to dock and re-enter first ship after argument function calls, as game client would guarantee next ship to be the first. It's hard to detect swiping to a _different_ ship, because users can have duplicate ships having same UI appearance.
>        Check the implementation of awakening: https://github.com/LmeSzinc/AzurLaneAutoScript/blob/aa12f86ea91764909a8d82b4bc6ae3a6e8dcbc44/module/awaken/awaken.py#L340-L364

I understand that, however compared to awakening, the ships that have gotten their augments upgraded to the max, don't get removed from the in game filter like awakening does, so that isn't really a solution either.
I would have assumed that if the module is empty (as it probably would be if there were duplicate ships) it would move on to the next ship already, since it doesn't fullfill the requirement of having an augment equipped.

PR #4603, opened 2025-02-23
EXP_INFO_B asset fix

Shane-Xue:
* Redo asset: EXP_INFO_B: This asset has repeatedly caused issues that had to be fixed. Now Let's fix it once and for all.

* Remove Previous Fix: Previously I checked for the asset of Chapter name, now we don't need it

* Tweak Base class name: Incorrect naming convention.

I don't know why the asset size is now smaller. But it runs on my local machine and let's just pretend everything's fine.


LmeSzinc:
`BATTLE_STATUS_*` 和 `EXP_INFO_*` 误判是因为它们都是平均颜色识别，在随机背景下就有可能会产生，把它们的识别位置换到另一个位置并不会从根本解决问题，可能这个移动解决了你关注的问题A，但是也可能产生问题BCD

Shane-Xue:
When rebasing this asset what I had in mind was the main storyline background, which has the title box (a gray color background) on the new asset I uploaded. On second thought, this is not the only entrance to combats, though even so I do not see any reason to believe this tweak will cause further issues.

Issue #4526, opened 2025-01-19
Feature Event end task and Meta reward


dragonheart107:
### Is your feature request related to a problem?

For the Event End Task, can Gacha task be included in the set of changed tasks?
But instead of disabling like other event tasks, set it to light and 1 pull if the event has ended?

For Meta Rewards:
There is the possibility of the following situation while META BATTLES' option One Hit Mode is selected and not having an actual 'oneshot-fleet'. This potentially could lead to missing out point rewards:

your own Meta battle: use assist option, but not yet finished -> task delayed -> missing out on today's point for now
ALAS does assist meta bosses from others-> finishes-> checks for reward-> no reward yet
ALAS checks back after delay for your own META-boss-> receive points -> new reward score reached
doesn't get checked though until next days META task as METAreward is tied to META assist
![Image](https://github.com/user-attachments/assets/8fec84bd-50fe-4d44-924d-241d339a535b)

I'm unsure where to start in order to  implement this, so just a feature request for now.

### Describe the solution you'd like

Event end task additionally changes gacha pool and amount to light and 1.

META reward gets checked not only after META assist, but also when META BATTLE task is finished / set to the next server refresh.

PR #4633, opened 2025-03-02
Feat: Event General End Time based on server time + modify Gacha task config on pool switch

dragonheart107:
Partial feature of #4526

this should allow for the input of the in game notification for the event ending timer and additionally will not disable gacha task and instead modifies it to 1 pull from light.

However if Gacha task is triggered before the other event tasks it will do the gacha as normal without stopping and event time limit will only be triggered if the other event tasks are run. How/where would I need to modify, so that it already stops on gacha too and does the disable task chain.


LmeSzinc:
1. EventGeneral_TimeLimit is a manual input value, it is supposed to be a local time to prevent calculations in human brain
2. Current gacha task will auto fallback to light pool if event is unavailable, which I think better as it's a once-and-done setting, user don't bother changing it every event

dragonheart107:
> 1. EventGeneral_TimeLimit is a manual input value, it is supposed to be a local time to prevent calculations in human brain
> 
>     2. Current gacha task will auto fallback to light pool if event is unavailable, which I think better as it's a once-and-done setting, user don't bother changing it every event

for 1. this is only true though if you are playing within the same time zone of your server, where local = server time; however while this is largely true for cn / jp players globals has too many different timezones; with the change you now need to only input the time mentioned in the in game announcements  when a event is ending, as ALAS then checks if the server time has been exceeded by local time, no need for calculations anymore

for 2. i guess that would be fine, however the amount of pulls doesn't get changed at the moment, no?

dragonheart107:
```
  if target_pool == 'wishing_well':
            if self._gacha_side_navbar.get_total(main=self) != 5:
                logger.warning('\'wishing_well\' is not available, '
                               'default to \'light\' pool')
                target_pool = 'light'
                #self.config.cross_set(keys=f'Gacha.Gacha.Amount', value= 1)
                self.config.__setattr__(value=1)
 ```
 would this work if added to `gacha-goto-pool` to not only default to light but also set pull amount to one?

dragonheart107:
unsure if there is another way aside from the config cross set, but it works now as I intended to:

1. Event time: only need to input the event end time from announcement (no change for JP or CN players since local = server time), for EN no need to calculate beforehand what the server time would be in local time

2. if gacha goes through the target pool loop, it also adjusts the pool and amount in config:

old: set to event 3x but no event pool -> pulls from light 3x and keeps config at event 3x
now: set to event 3x but no event pool -> pulls from light 1x and updates config to light 1x

dragonheart107:
what do you think? @LmeSzinc

dragonheart107:
hm, guess I'll just keep this for my instance then

PR #4672, opened 2025-03-19
Various asset related changes

dragonheart107:
- BATTLE_STATUS assets having different buttons is outdated for EN since whole screen counts to advance
- Upd: statistics assets for en
- ~~Feat: OCR detection for EN server only, otherwise use previous logic~~
- Fix: missing import for item_statistics
- Fix: handle the special D screen outside of exercise too

Edit: actually unsure if the ocr works


module/statistics/campaign_bonus.py
```diff
@@ -13,12 +16,28 @@ def predict_valid(self):
...
+    def ocr_object(self):
+        return Ocr(lang='cnocr', threshold=128, name='REWARDS_OCR')
+
     def appear_on(self, image):
-        if AUTO_SEARCH_MENU_EXIT.match(image, offset=(200, 20)) \
-                and CAMPAIGN_BONUS.match(image, offset=(20, 500)):
-            return True
+        #Unsure if other servers need an updated version too, or if previous worked
+        if server.server == 'en':
+            if not AUTO_SEARCH_MENU_EXIT.match(image, offset=(200, 20)):
```
dragonheart107, line 26: here, i changed this to, since I couldn't get not match and match to give me the correct value:
```
            if AUTO_SEARCH_MENU_EXIT.appear_on(image):
```
it actually recognizes it, but why does it not work with match?

module/combat/combat.py
```diff
@@ -373,8 +373,13 @@ def handle_battle_status(self, drop=None):
...
                 self.device.sleep((0.25, 0.5))
             self.device.click(BATTLE_STATUS_D)
-            return True
 
+            if self.appear(OPTS_INFO_D, interval=self.battle_status_click_interval):
+                self.device.click(OPTS_INFO_D)               
+            # Click BATTLE_STATUS_D again
+            self.device.click(BATTLE_STATUS_D)
+            return True
+
```
LmeSzinc, line 382: Handling D screen isn't just clicking OPTS_INFO_D, there are a lot of post process to do, like reduce mood, decide to continue with another fleet or retreat, count as task failure

assets/en/combat/BATTLE_STATUS_A.BUTTON.png
LmeSzinc, line 1: Keep `.BUTTON` assets. `BATTLE_STATUS_*`, `GET_ITEMS_*`, `EXP_INFO_*` share clicking area to have continous clicks after each battle, and the area is a safe-to-click area that has no conflict with other buttons like SWITCH_OVER

module/statistics/campaign_bonus.py
```diff
@@ -13,12 +16,21 @@ def predict_valid(self):
...
+        #Unsure if other servers need an updated version too, or if previous worked
+        if server.server == 'en':
+            if AUTO_SEARCH_MENU_EXIT.appear_on(image):
+                ocr_result = self.ocr_object.ocr(image)
+            return "Rewards" in ocr_result
+        else:
+            if AUTO_SEARCH_MENU_EXIT.match(image, offset=(200, 20)) \
+                    and CAMPAIGN_BONUS.match(image, offset=(20, 500)):
+                return True
+            return False
```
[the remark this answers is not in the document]
dragonheart107: yep, fixed this, I forgot to set server to en when testing, so it was ofc using the cn asset and wouldn't match

assets/en/combat/BATTLE_STATUS_A.BUTTON.png
dragonheart107: alright

module/combat/combat.py
dragonheart107: what do you think should happen if fleet gets defeated:

-abandon task(retreat, set config to false etc), stop alas
-abandon task, keep going with other tasks
-try task one more time, if fail again remove task
-let user decide with a new GUI option

PR #4756, opened 2025-04-25
Add: Private Quarters support

nEEtdo0d:
Stable for EN

- Purchases weekly roses, only buys all or none.  Cheap enough price that account should have 24K easy especially when using ALAS
- Executes interact sequence with target ship girl
- Assumes necessary resources are downloaded, if detects pop up then exits
- Retries in case ship girl doesn't appear in initial load, fixed number of retries, only Sirius is known to do this

Besides EN, servers should replace their respective assets, pretty much anything with english text and update their language json.  Can opt in using other common markers in a page instead to be more flexible with other servers.  The assets are catered for EN.

Will merge / rebase as needed.


assets/cn/private_quarters/PRIVATE_QUARTERS_ROOM_TARGET_CLICK_AREA.png
LmeSzinc, line 1: paint content white to reduce file size as image content is not needed here

module/private_quarters/private_quarters.py
```diff
@@ -0,0 +1,289 @@
...
+        """
+        Callable wrapper to validate target's appearance
+        offset=(100, 100) detectable for anchorage, noshiro, and sirus
+        When more ships added may need to adjust or capture specific bubble position per
+        ship, can use the available_targets to store similarly into tuples instead
+        """
+        settle_timer = Timer(1.5, count=3).start()
+        settle_timer.wait()
+        self.device.screenshot()
+        return self.appear(PRIVATE_QUARTERS_ROOM_TARGET_CHECK, offset=(100, 100))
```
LmeSzinc, line 35: wait with image detection, don't use plain wait

module/private_quarters/private_quarters.py
```diff
@@ -0,0 +1,289 @@
...
+            check_button=self._pq_goto_room_check,
+            appear_button=page_private_quarters.check_button,
+            offset=(20, 20),
+            skip_first_screenshot=False)
+
+        # If was download asset popup
+        # Terminate the run
+        if self.handle_popup_cancel('PRIVATE_QUARTERS_DOWNLOAD_ASSET', offset=(20, 20)):
+            logger.error(f'Cannot enter {target_title}\'s room, please download the necessary assets first')
+            return False
```
LmeSzinc, line 69: Function with device operation cannot be end condition

module/private_quarters/private_quarters.py
```diff
@@ -0,0 +1,289 @@
...
+        # Ensure either loading or popup
+        # prompt appears after click
+        target_title = target_ship.title()
+        target_btn = self.available_targets[target_ship]
+        self.ui_click(
+            click_button=target_btn,
+            check_button=self._pq_goto_room_check,
+            appear_button=page_private_quarters.check_button,
+            offset=(20, 20),
+            skip_first_screenshot=False)
```
LmeSzinc, line 63: always use skip_first_screenshot=True

module/private_quarters/private_quarters.py
```diff
@@ -0,0 +1,289 @@
...
+            if self.appear(PRIVATE_QUARTERS_ROOM_CHECK, offset=(20, 20)):
+                break
+
+            # Continue without clicking, mitigate too many click exception
+            if self.appear(PRIVATE_QUARTERS_LOADING_CHECK, offset=(20, 20)):
+                continue
+
+            if click_timer.reached():
+                self.device.click(PRIVATE_QUARTERS_ROOM_SAFE_CLICK_AREA)
+                click_timer.reset()
```
LmeSzinc, line 91: clicking need to be triggered by `appear(button)`

module/private_quarters/private_quarters.py
```diff
@@ -0,0 +1,289 @@
...
+                self.device.click(PRIVATE_QUARTERS_ROOM_TARGET_CLICK_AREA)
+                click_timer.reset()
+            if self.appear_then_click(PRIVATE_QUARTERS_INTERACT, offset=(20, 20), interval=1):
+                click_timer.reset()
+            if self.appear(PRIVATE_QUARTERS_INTERACT_CHECK, offset=(20, 20), interval=1):
+                self.device.click(PRIVATE_QUARTERS_ROOM_BACK)
+                click_timer.reset()
+                click_iteration += 1
+
+        self._pq_goto_room_exit()
```
LmeSzinc, line 206: End condition must be image detection

assets/cn/private_quarters/PRIVATE_QUARTERS_ROOM_TARGET_CLICK_AREA.png
nEEtdo0d: good point, will do

module/private_quarters/private_quarters.py
nEEtdo0d: do you mean use a regular while loop with screen shot arrangement or is there a module function that can help with this?

the reason for the wait is because the ship girl is not guaranteed to appear on room enter, sirius still does this, and anchorage jumps around so will mistake for the ship girl not being there

didn't use a while loop since needed reasons for it to return true or false

otherwise using the room as static elements but would require capturing image for each ship girl's room, seemed overkill in that sense

module/private_quarters/private_quarters.py
nEEtdo0d: not sure I understand this, once cancelled there isn't any other operation to perform so just have to exit the task altogether
the size of the download can vary so thought it was best for the user to address the asset download themselves

also there aren't great indicators of the download being in progress either

module/private_quarters/private_quarters.py
nEEtdo0d: this might be reasonable, I'll test it if I recall the image it has on hand at the time should be sufficient for this operation of ui_click

module/private_quarters/private_quarters.py
nEEtdo0d: after loading, the ship girl will talk and there aren't any UI elements to really verify or validate on so have to rely on clicking a random space to progress the conversation and fully enter the room

the ship girl themselves is 3d so can't really count on a static image from it

module/private_quarters/private_quarters.py
nEEtdo0d: would combination of both be acceptable, I think this might work
`if click_iteration > 2 and self.appear(PRIVATE_QUARTERS_INTERACT)`

to complete the task, needs cycle in this loop at least 3 times

assets/cn/private_quarters/PRIVATE_QUARTERS_ROOM_TARGET_CLICK_AREA.png
nEEtdo0d: done in 3eb37033a17bb9af73fb16d984e61ca089e2f6a3

module/private_quarters/private_quarters.py
nEEtdo0d: done in 3eb37033a17bb9af73fb16d984e61ca089e2f6a3

module/private_quarters/private_quarters.py
nEEtdo0d: done in 3eb37033a17bb9af73fb16d984e61ca089e2f6a3

hopefully acceptable, ill keep this comment unresolved if unsatisfactory

module/private_quarters/private_quarters.py
nEEtdo0d: following up, still not sure I understand this

any suggestions on how to refactor?

Like said, the function is intended to enter into a room, either able to with loading screen or the normal confirm/cancel pop up

module/private_quarters/private_quarters.py
nEEtdo0d: following up, after the loading screen besides the ship girl is the only UI element in the shot besides her personal room belongings

But having to capture each girl's specific room doesn't seem productive when the idea is to progress forward to get to the actual UI of the private quarters feature.

PR #4813, opened 2025-05-20
Fix: adapt new scrap UI for JP and fix EN

dragonheart107:
~~TODO other servers~~

`self.interval_clear(GET_ITEMS_1)`
unsure about above part with the scrap change, wether it needs to be moved or not


LmeSzinc:
check if #4818 works

dragonheart107:
seems like it now uses the correct assets

PR #4850, opened 2025-05-30
Fix: asset change after patch

dragonheart107:
tactical class ship selection got a menu change/reposition with patch, check if other servers need it too


LmeSzinc:
no changes on CN/TW

PR #4866, opened 2025-06-06
Fix: Meta-sync asset change for EN

dragonheart107:


dragonheart107:
actually wait, don't merge yet, it seems they use 2 different ones depending on what stage of sync it is at

PR #4907, opened 2025-06-23
Fix: dropscreen for autosearch reward screens

dragonheart107:


dragonheart107:
also, I noticed that hard mode stage 13-4 is saved as 12-4:
grafik

dragonheart107:
any feedback @LmeSzinc ?

PR #4944, opened 2025-07-10
Fix: Private Quarters weekly rose purchase logic

ChthoIly:
修复https://github.com/LmeSzinc/AzurLaneAutoScript/issues/4943#issue-3220411769 的问题


module/private_quarters/private_quarters.py
```diff
@@ -238,6 +238,29 @@ def pq_shop_weekly_roses(self):
             if self.appear(PRIVATE_QUARTERS_SHOP_WEEKLY_ROSES_CHECK, offset=(20, 20)):
                 break
```
ChthoIly, line 239: 我不太确定，但是这一段应该有检查浪漫满分的购买情况的功能。但是我在国服我的账号上时遇到了https://github.com/LmeSzinc/AzurLaneAutoScript/issues/4943#issue-3220411769 中提到的问题

我不确定 @nEEtdo0d 宿舍计划原作者处理这个问题的本意是什么样的，可能是CN服和EN服UI逻辑有不一样的地方？

module/ui/page.py
```diff
@@ -269,7 +269,7 @@ def link(self, button, destination):
 # Private Quarters
 page_private_quarters = Page(PRIVATE_QUARTERS_CHECK)
 page_dormmenu.link(button=DORMMENU_GOTO_PRIVATE_QUARTERS, destination=page_private_quarters)
-page_private_quarters.link(button=GOTO_MAIN_WHITE, destination=page_main)
+page_private_quarters.link(button=PRIVATE_QUARTERS_GOTO_MAIN, destination=page_main)
```
ChthoIly, line 272: 好像宿舍计划的UI布局更新了，在home键旁边加了个去设置的按钮导致旧资源回不到主界面而是进入设置后无法识别而报错停止

需要确认外服是否与国服一起更新了UI布局，如果没有可能导致其他服务器在宿舍计划界面无法返回主界面

@nEEtdo0d Could you confirm whether EN server has also updated the home button layout for private quarters?
 CN server looks like this now:
 
image

ChthoIly:
目前改动后国服测试宿舍计划已可以正常运行

module/private_quarters/private_quarters.py
nEEtdo0d: Looks to be a difference between CN and EN, https://imgur.com/a/JaTtJg0

After making the weekly purchase, all subsequent entering of the shop would like this in EN and I guess CN approaches it differently

I do recall feeling I've run into this one time but after ALAS restarts and retries it goes back to the image I have posted

Hence for EN is sufficient that the rose icon is just outright missing in that case.

I think the way to address this is by separate the function `_pq_shop_weekly_roses` for the server, hopefully this gets the idea across
https://pastebin.com/PEiS5wP2

because I am unable to replicate a similar image you have it now, it just advertises if sold out displays a new item instead onto the shop.

until I am able to get sample image similar to CN, I would think this is the best way to do for less confusion

module/ui/page.py
ChthoIly: With the original asset there are high chance which it will click on the settings area and cause the program to lost track and throw exception.

I am not sure if replacing `GOTO_MAIN_WHITE` might cause problem or not. Some other task may depend on the old asset's coordinates? As what I can see only private quarters has the home and settings button sharing the right up conner like this layout.

assets/cn/private_quarters/PRIVATE_QUARTERS_PAGE_BACK.png
nEEtdo0d, line 1: Asset seems to be not in use, can delete?

Besides I think already exists https://github.com/LmeSzinc/AzurLaneAutoScript/blob/master/assets/cn/private_quarters/PRIVATE_QUARTERS_SHOP_BACK.png

module/private_quarters/private_quarters.py
ChthoIly: Now seeing you picture it came across to me this might be because I have bought all the `Cake Gift` for this week. Maybe that's why `Roses` does not show on your end when you bought all of it.

I can't verify this as I have to wait till next week to have the rose and cake refreshed.

But there is a UI bug in CN that I can confirm which is after you buy all the roses and restart the game the rose icon will not prompt `sold out` and still clickable and brings you to the purchase page. Is just you cannot actually buy it. 

You can see that from https://github.com/LmeSzinc/AzurLaneAutoScript/issues/4943#issue-3220411769 the last video shows that. The UI goes back to normal after you exit the shop once and opened it again. The same issue will happen if the game is restarted.

I will keep looking into this issue. I am also thinking of adding the option to buy `Cake Gift` if there are people want it.

nEEtdo0d:
If not too much trouble, I think bump this up is acceptable

since it appears 1.5 second retry might be too fast and presses too quickly, maybe to at least 3 second
https://github.com/LmeSzinc/AzurLaneAutoScript/blob/5682daccd0f91c2739344f05efd38468b97ef979/module/private_quarters/private_quarters.py#L188C13-L188C23

didn't happen often before but all of sudden now is haha, so it presses that button too quickly and as a result ends up in `page_main or page_main_white`, thereby causing the below exception

```
2025-07-09 08:39:35.278 | INFO | INTERACT END                                                                          
2025-07-09 08:39:35.280 | INFO | <<< UI CLICK >>>                                                                      
2025-07-09 08:39:35.298 | INFO | Click ( 140,   41) @ PRIVATE_QUARTERS_ROOM_BACK                                       
2025-07-09 08:39:36.901 | INFO | Click (  95,   43) @ PRIVATE_QUARTERS_ROOM_BACK                                       
2025-07-09 08:39:38.514 | INFO | Click ( 104,   48) @ PRIVATE_QUARTERS_ROOM_BACK                                       
2025-07-09 08:40:38.762 | INFO | Function calls:                                                                       
           <string>   L1 <module>                                                                                      
           spawn.py L105 spawn_main()                                                                                  
           spawn.py L118 _main()                                                                                       
         process.py L297 _bootstrap()                                                                                  
         process.py  L99 run()                                                                                         
 process_manager.py L166 run_process()                                                                                 
            alas.py L558 loop()                                                                                        
            alas.py  L69 run()                                                                                         
            alas.py L242 private_quarters()                                                                            
private_quarters.py L418 run()                                                                                         
private_quarters.py L403 pq_run()                                                                                      
private_quarters.py L327 pq_interact()                                                                                 
private_quarters.py L182 _pq_goto_room_exit()                                                                          
              ui.py L108 ui_click()                                                                                    
          device.py L184 screenshot()                                                                                  
          device.py L246 stuck_record_check()                                                                          
          device.py  L43 show_function_call()                                                                          
2025-07-09 08:40:38.768 | WARNING | Wait too long                                                                      
2025-07-09 08:40:38.771 | WARNING | Waiting for {'PRIVATE_QUARTERS_CHECK', 'PRIVATE_QUARTERS_ROOM_BACK'}               
2025-07-09 08:40:38.809 | INFO | [Package_name] com.YoStarEN.AzurLane                                                  
2025-07-09 08:40:38.811 | ERROR | GameStuckError: Wait too long                                                   
```

module/private_quarters/private_quarters.py
nEEtdo0d: please do, I don't think I've come across that `sold out` scenario

but this is why you have the `PRIVATE_QUARTERS_SHOP_WEEKLY_ROSES_SOLD_OUT` correct? so the 2nd pastebin should work out too I think, give it a try if you can, it should not click the button either I think?

`Cake gift` is luxury I cannot afford as F2P haha but adding the option is fine I will definitely have it as `false`

module/ui/page.py
ChthoIly: with GOTO_MAIN_WHITE as the target action it will just click on the settings button. You can see this from the coordinate diff:
PRIVATE_QUARTERS_GOTO_MAIN
GOTO_MAIN_WHITE

They don't overlap at all.

ChthoIly:
> since it appears 1.5 second retry might be too fast and presses too quickly, maybe to at least 3 second

this did not come across me, room back has always worked for me. but it couldn't hurt to increase a little wait time.

module/private_quarters/private_quarters.py
ChthoIly: > but this is why you have the `PRIVATE_QUARTERS_SHOP_WEEKLY_ROSES_SOLD_OUT`

Yes. because is hard to figure out a precise way to check for `sold out` with the UI bug issue. So I just decided to simplify enter the shop and exit and enter it again to bypass the UI bug. The second entering will make sure the rose's status is correct.

Maybe after next week I confirmed the shop list ordering issue or someone confirmed to have the same issue. We can find a way to optimize the code to make it uniform.



> `Cake gift` is luxury I cannot afford as F2P haha

I always clear out the cake so the intimacy goes up quickly. But just as other gem buying option like the daily shop refresh, it should be false by default.

module/ui/page.py
nEEtdo0d: true they do not overlap, I guess the `offset` is not cutting it anymore looks like I am experiencing this too, I guess the new update with research s8 out they made some new GUI adjustments too so better to use a new asset altogether specific for private quarters
```
INFO     20:33:42.435 │ Scheduler: End task `PrivateQuarters`                      
INFO     20:33:42.439 │ [Server] en                                                
INFO     20:33:42.478 │ Bind task ['General', 'Alas']                              
INFO     20:33:42.487 │ Pending tasks: ['OpsiAshBeacon', 'OpsiDaily', 'OpsiShop',  
         'OpsiAbyssal', 'OpsiStronghold', 'OpsiObscure', 'Daily', 'Hard',          
         'OpsiAshAssist', 'Main', 'OpsiMeowfficerFarming']                         
INFO     20:33:42.490 │ [Task] OpsiAshBeacon (Enable, 2025-07-10 03:00:00)         
INFO     20:33:42.493 │ Bind task ['General', 'Alas', 'OpsiGeneral',               
         'OpsiAshBeacon']                                                          
INFO     20:33:42.528 │ Scheduler: Start task `OpsiAshBeacon`                      
═══════════════════════════════════════════════════════════════════════════════════
                                   OPSIASHBEACON                                   
═══════════════════════════════════════════════════════════════════════════════════
INFO     20:33:43.020 │ <<< UI ENSURE >>>                                          
INFO     20:33:43.023 │ UI get current page                                        
INFO     20:33:43.105 │ [UI] page_private_quarters                                 
INFO     20:33:43.113 │ Goto page_reward                                           
INFO     20:33:43.120 │ <<< UI GOTO PAGE_REWARD >>>                                
INFO     20:33:43.164 │ Page switch: page_private_quarters -> page_main            
INFO     20:33:43.170 │ Click (1229,   31) @ GOTO_MAIN_WHITE                       
INFO     20:33:48.517 │ Page switch: page_private_quarters -> page_main            
INFO     20:33:48.522 │ Click (1229,   37) @ GOTO_MAIN_WHITE                       
INFO     20:33:53.664 │ Page switch: page_private_quarters -> page_main            
INFO     20:33:53.668 │ Click (1228,   44) @ GOTO_MAIN_WHITE                       
INFO     20:33:58.854 │ Page switch: page_private_quarters -> page_main            
INFO     20:33:58.858 │ Click (1224,   30) @ GOTO_MAIN_WHITE             
```

like said I think your new button works a for EN too but if it does not I can address it. Or if you do not mind, you can replace it in the assets/en folder as part of this PR since you do have access to a sample image from one of my earlier links.

dragonheart107:
Regarding private quarter to main, see PR #4942, isn't that already the solution for that specific issue?

Also I got the rose UI glitch on EN now too, though the only indicator that roses for the week are already bought is the counter in Pop-Up being 8/8 already
MuMu12-20250711-163530
MuMu12-20250711-163540
MuMu12-20250711-163543
what i noticed: complete restart (close AL then start up) makes it reproduceable, while if I just go out from PQ into Main then reenter PQ it fixes itself

ChthoIly:
> Regarding private quarter to main, see PR #4942, isn't that already the solution for that specific issue?

Yes, is basically the same fix, I will be removing that. I noticed you didn't add `GOTO_HOME` assets for EN/JP/TW servers. I will make changes to the `GOTO_HOME` assets name and place them in others servers assets file. From what I can see, it seems CN/EN server has the same layout for the button so assuming other server is the same too.

> Also I got the rose UI glitch on EN now too

Did you bought all the `Cake Gift` this week? Its seems this only happens when `Cake Gift` is sold out. For the people who didn't buy all the `Cake Gift`, rose should not appear on the page. This is also what the original code is doing, checking is the rose is missing for availability.

ChthoIly:
不是很会用git，这里commit搞砸了。我关掉重新弄一个

dragonheart107:
> > Regarding private quarter to main, see PR #4942, isn't that already the solution for that specific issue?
> 
> Yes, is basically the same fix, I will be removing that. I noticed you didn't add `GOTO_HOME` assets for EN/JP/TW servers. I will make changes to the `GOTO_HOME` assets name and place them in others servers assets file. From what I can see, it seems CN/EN server has the same layout for the button so assuming other server is the same too.
> 
> > Also I got the rose UI glitch on EN now too
> 
> Did you bought all the `Cake Gift` this week? Its seems this only happens when `Cake Gift` is sold out. For the people who didn't buy all the `Cake Gift`, rose should not appear on the page. This is also what the original code is doing, checking is the rose is missing for availability.

I didn't put assets of the other servers for `GOTO_HOME` assets on purpose as by ALAS default rules it will use cn asset ones if other servers don't use their own ones, to save some file size

for second part: yes i bought the weeklies already, but UI glitch happened. compared to CN though, it won't have the bought out hanzi (i assume that is what they say) instead it shows just like normal, only in the pop-up after clicking on the roses, it will show (Weekly purchase 8/8), so EN would need to work with this trigger

ChthoIly:
> it won't have the bought out hanzi (i assume that is what they say) instead it shows just like normal

have you tried to reenter the shop again to see if the sold out appears? reentering the shop will fix the glitch and you either no longer see the item (in cake gift's case) or it will display sold out on the item and you will no longer able to open the purchase page by clicking.

It is very hard for me to test on other servers, since private quarters is only available to player level >80. I don't have account on hand thats >80 lvl in EN server. Even CN server if I want to test for cake gift purchase option. I have to wait till next week when my account refresh. I don't have other accounts to test with...

dragonheart107:
> > it won't have the bought out hanzi (i assume that is what they say) instead it shows just like normal
> 
> have you tried to reenter the shop again to see if the sold out appears? reentering the shop will fix the glitch and you either no longer see the item (in cake gift's case) or it will display sold out on the item and you will no longer able to open the purchase page by clicking.
> 
yes this is consistent for me too
from first app start: -> bugged UI -> fixed by you exit private quarters to the main screen and then reenter without logging out/closing and restarting the game, however not the bought out, but the 'next' item, cake for example
MuMu12-20250712-020946



also think it's better to chat in the new PR from now on

PR #4949, opened 2025-07-11
Fix: Private Quarters weekly rose purchase logic

ChthoIly:
Fix: increased _pq_goto_room_exit() retry wait time
Add: [EN][JP][TW] Private Quarter GOTO MAIN asset update


ChthoIly:
重新开的https://github.com/LmeSzinc/AzurLaneAutoScript/pull/4944#issue-3220616482

nEEtdo0d:
Sorry just now catching up, thank you @dragonheart107 for the sample images for EN from https://github.com/LmeSzinc/AzurLaneAutoScript/pull/4944#issuecomment-3061168500 and also fixing the `goto_main` bug, newest game update must have done this since it was definitely by itself before heh.

I cannot get the UI glitch to appear for me even when doing either of the methods suggested. I still get the cake.  @ChthoIly and I thought it might be because accounts who bought the cake using gems might be the trigger, @dragonheart107 confirm whether this is true? Or you got this even without making any purchase to cake as well? What happens if you try to make the purchase anyway, do info bars show up? Or just no respone and stays on the purchase page?

Cannot say I like the idea of exiting and going back into the private quarters page but if its the only solution that works maybe have to then. A similar example I remember is the Operation Siren fog status can always follow you, only way I know is have to terminate the whole app and relaunch for it to go away.

Preferably assuming the same scenario applies to other servers like @dragonheart107 sample images, edit the purchase loop a small bit. Or using a timer as another exit clause besides seeing `PRIVATE_QUARTERS_SHOP_WEEKLY_ROSES_GET`
https://pastebin.com/3Nbk1Tr0

Still think this is good to have as well with @ChthoIly sample image showing that in the shop page just in case.
https://pastebin.com/wJ5xNQZC

Using OCR on the weekly number in the purchase page may differ across servers or might get blocked by the coin image.  So having to capture a unique image per server would be annoying.  Ideally a solution applicable to all servers is preferred. If the number in the shop page actually updated that might be okay but the sample image says otherwise unfortunately.

If the exit and enter shop is the only way to address this, I would opt in this arrangement
https://pastebin.com/yvWhjxre
Combine with https://pastebin.com/wJ5xNQZC
More or less the same thing you have now just doing it earlier and no server restriction.

nEEtdo0d:
A completely different option is to use the other shop pages to make the purchase rather than using the front page.  Would hope those are at least accurate, might have to go this route one day anyway since purchasing cake is a future enhancement.
Click on Sirius (or any ship girl I think) image, scroll horizontally to get to the roses and cake area, read weekly count there, and make the purchase there. The scroll would be the annoying part though.

Otherwise, to make cake purchase might be just to again do another enter-exit shop until it shows up on the front page and similarly do the same purchase logic we have now.

dragonheart107:
>@ChthoIly and I thought it might be because accounts who bought the cake using gems might be the trigger, @dragonheart107 confirm whether this is true? Or you got this even without making any purchase to cake as well?

never bought the cakes

>What happens if you try to make the purchase anyway, do info bars show up? Or just no respone and stays on the purchase page?

for cake or rose?

dragonheart107:
also while not ideal, manjuu might potentially fix this

ChthoIly:
Any refresh in UI page will get the glitch fixed. whether reentering shop or click on any ship girl in the shop and back.

https://github.com/user-attachments/assets/d5bff0ea-5829-4ab5-97b6-4e32db5ed99d

I can change the reentering shop logic to switch to a ship girl and switch back, but they seem to take the same time to load. No really time improve on this. @nEEtdo0d do you think is necessary to change?

dragonheart107:
the loadtime difference might be bigger for 'worse' devices for main into private quarters comapred to shop main to shipgirl and back to shop main, though I don't really have a way to test this myself, I feel like if no users complain that much keep it as is?

ChthoIly:
I feel so confused. How can @nEEtdo0d don't get this glitch but @dragonheart107 do? Both of you are on EN server right? and DID NOT bought any cakes? I thought the glitch should be consistence.

From what I have tested on CN server if:

first start app -> glitch UI -> exit shop or switch to a ship girl within the shop -> back to the main shop page -> glitch fixed -> from this point on the glitch will never show again unless restart the app

This is very consistence for me, 100% reproducible.

image
image

> the loadtime difference might be bigger for 'worse' devices for main into private quarters comapred to shop main to shipgirl and back to shop main, though I don't really have a way to test this myself, I feel like if no users complain that much keep it as is?

Couldn't be hard to change the refresh logic to switch ship girls. Just a quick implementation. This way the entire process happens within the shop, so maybe it should be a little faster?

dragonheart107:
Yeah really weird that for me it's exactly as you say, but not for him. Maybe different servers (mine is Lexington) on the EN server behave differently?

ChthoIly:
I am assuming the default shop listing logic is:

- show rose first over cake
- show cake if rose are sold out
- show rose UI glitch if both rose and cake are sold out

this is important to be true for the cake purchase option that I am doing now.

I can't verify this as I don't have another account which has both items available. Have to wait till next Monday to check.

nEEtdo0d:
> for cake or rose?

@dragonheart107 for rose in the images you shared that was when you already made the purchase. So curious what happens if you try to still make the purchase in this window.

> I feel so confused. How can @nEEtdo0d don't get this glitch but @dragonheart107 do? Both of you are on EN server right? and DID NOT bought any cakes? I thought the glitch should be consistence.

> Yeah really weird that for me it's exactly as you say, but not for him. Maybe different servers (mine is Lexington) on the EN server behave differently?

Avora and Amagi for me.  I think it might be difference in emulator if anything. I use BlueStacks, been seeing people prefer over using Mumu nowadays? BlueStacks is not perfect but too lazy to try another haha.  Couldn't get Mumu to work properly, used Nox before but got worst performance over time.

> first start app -> glitch UI -> exit shop or switch to a ship girl within the shop -> back to the main shop page -> glitch fixed -> from this point on the glitch will never show again unless restart the app

> Couldn't be hard to change the refresh logic to switch ship girls. Just a quick implementation. This way the entire process happens within the shop, so maybe it should be a little faster?

Like the idea better than exit and enter shop. Stay in the shop GUI this way. If consistent and will properly update to reflect no roses are available then cool with me! @ChthoIly Good discovery finding that is an option. Yeah, more streamline and faster to confirm, not too much loading in comparison to exit and enter shop I think.
Agreed should be fairly straight forward to add, for cleanliness make it under another def like `_pq_shop_select()`

If want to go the extra mile, have it with argument for streamline
```
self._pq_shop_enter()

# address UI glitch bug
self._pq_shop_select('sirius') # sirius page
self._pq_shop_select('home') # front page

# same waiting mechanism to ensure roses are there and what not
```

Another alternative.  If know how to use, wrap the select using this class. https://github.com/LmeSzinc/AzurLaneAutoScript/blob/master/module/ui/navbar.py
There should be plenty of examples in code you can reference and try out with. Works for both horizontal and vertical, just need to set it up correctly.  It is what is used for the ship enhancement and shop feature.

nEEtdo0d:
> I am assuming the default shop listing logic is:
> 
> * show rose first over cake
> * show cake if rose are sold out
> * show rose UI glitch if both rose and cake are sold out
> 
> this is important to be true for the cake purchase option that I am doing now.
> 
> I can't verify this as I don't have another account which has both items available. Have to wait till next Monday to check.

Looks right except @dragonheart107 said he did not make any purchase to cake. Actually he has a different GUI but hopefully the switching between shop front page to shop ship girl page to shop front page will remedy for all server.

UI glitch is not guaranteed either seeing as I've yet to come across it even doing the same steps as described

dragonheart107:
>So curious what happens if you try to still make the purchase in this window.

closes 
MuMu12-20250711-163543
and shows this error
MuMu12-20250712-020844

> > I feel so confused. How can @nEEtdo0d don't get this glitch but @dragonheart107 do? Both of you are on EN server right? and DID NOT bought any cakes? I thought the glitch should be consistence.
> 
> > Yeah really weird that for me it's exactly as you say, but not for him. Maybe different servers (mine is Lexington) on the EN server behave differently?
> 
> Avora and Amagi for me. I think it might be difference in emulator if anything. I use BlueStacks, been seeing people prefer over using Mumu nowadays? BlueStacks is not perfect but too lazy to try another haha. Couldn't get Mumu to work properly, used Nox before but got worst performance over time.

emulator might be the culprit then? using mumu12

>switching between shop front page to shop ship girl page to shop front page will remedy for all server.

can confirm that it does fix the glitch for me too

ChthoIly:
> So curious what happens if you try to still make the purchase in this window.

For your curiosity, forcefully to make the purchase in this window when you actually has not more items left will prompt `System Error:40:Illegal limit condition`

Screenshot 2025-07-11 195846

I also see someone else in CN post video about the same glitch https://www.bilibili.com/video/BV1A3dqY4EHU/?vd_source=591c7a7918476704d065bf6f95ed78de


> I think it might be difference in emulator if anything

I don't really feels like this differ from emulators. I am able to reproduce this on my M2 iPad and iPhone. If native enviroment still has the issue then I would assume is Manjuu to blame ~~(They have like 0.5 programmers in total)~~. 



> can confirm that it does fix the glitch for me too

Good to hear. I will change to the switching ship girl logic then.

nEEtdo0d:
> I don't really feels like this differ from emulators. I am able to reproduce this on my M2 iPad and iPhone

Samsung Galaxy S20, I do not get it either. Maybe have to be really quick but yeah very strange.

ChthoIly:
Since, I am implementing `_pq_shop_select(self, shop_page)`. I think we can change purchasing and sold out check to this page by selecting on one of the ship girl and choose gift section:
image
This way the `_pq_shop_select(‘sirius’)` it self will fix the glitch. Therefore, we are not wasting any clicking step.


> Samsung Galaxy S20, I do not get it either. Maybe have to be really quick but yeah very strange.

I have a old Redmi Note 10 Pro, I have confirmed it happens on native android environment too. I would say is a bug from Manjuu. In your case maybe could be the serves difference? I mean both of you are on EN server so you should all have the same game file.

nEEtdo0d:
> Since, I am implementing `_pq_shop_select(self, shop_page)`. I think we can change purchasing and sold out check to this page by selecting on one of the ship girl and choose gift section:
> This way the `_pq_shop_select(‘sirius’)` it self will fix the glitch. Therefore, we are not wasting any clicking step.

Annoying part is the scroll if have to.  Maybe they add more items so rose in different spot later.  Sirius and Noshiro is the only one with 2 extra items while others have just 1. Though they are gifts, are they 1 time purchase weekly or forever? If the latter they  might get moved to the end thereby moving the cake and rose icons too. Or if they add more items for more special vignettes like Sirius.

Here are images for EN, if you need anything else let me know
https://imgur.com/a/en-shop-page-pq-shop-select-if-needed-jzBDA6f

I chose the front page because just focused on roses for F2P. First image on enter, same place every time, and no need to scroll.

ChthoIly:
I am thinking of using template matching to buy rose and cake in private quarters shop just like how is already been done in general shops. However, I am not sure how sold out is detected in general shops, I could use my own way to detect for sold out. But for code consistency if this has been done in general shop I would follow the same way.

I also thought of the scroll way like commission list but there isn't a distinct white line for identify in private quarters shop. Any advise for this, which method would be better? @nEEtdo0d

nEEtdo0d:
> I am thinking of using template matching to buy rose and cake in private quarters shop just like how is already been done in general shops

I think is probably a good direction to start in.  The annoying part general shop and most other shops have a distinct border around the item, but the items here do not.  But should still be fine more or less, choosing a good (x, y) coordinate, might take experimenting.  The items are still evenly spaced I think.
https://imgur.com/a/item-grid-visual-tjCERRx

A challenge might be how to detect the cost since the grid differs a lot from general shop.  The amount area is much farther away. `ItemGrid` I think is pretty flexible. I think it uses the (x, y) coordinate that for the item then adds +y to reach where the gem and number are.

> However, I am not sure how sold out is detected in general shops, I could use my own way to detect for sold out. But for code consistency if this has been done in general shop I would follow the same way.

Sold out items are greyed out and cover the item image.  The template matching assumes this is a new image and does not recognize it and will not add it to the return list of items.  So sold out items are basically ignored.
https://imgur.com/a/sold-out-VoaLJZX

Unfortunately, for the items here that is not the case. Rose despite being sold out is not greyed out and still in view but the cost is gone.  So the way to detect sold out items is probably 2 ways.
- On doing the buying step, attempt to buy it anyway and detect with `handle_infobar`.  So the ways to exit the buy step, if reached the purchase amount pop up or if the info bar is displayed.
- Have 2 `ButtonGrid or ItemGrid` one for the template matching the item and the other detecting the sold out word. Then lining these button grids up assuming they are equal in length and match one to one in position too. You may able to do this using 1 `ButtonGrid or ItemGrid` only but just a little harder because maybe drawing and checking for the position of sold out dynamically instead.

The first I think is the easiest, sure there is some wasted movement, but accomplishes the end goal just as good.

> I also thought of the scroll way like commission list but there isn't a distinct white line for identify in private quarters shop. Any advise for this, which method would be better?

Scroll is more challenge here because I think it relies on actual scroll bar being visible otherwise not able to tell whether its at the top or bottom of page.
https://imgur.com/a/scroll-bar-commission-shop-vs-private-quarters-1iwZR8x

There is no scroll bar for private quarters shop. For now scroll might have to be pushed to the side, just assume rose and cake are always in view using the gift page.

When it comes to detecting a dynamic `ButtonGrid or ItemGrid` the owner relies on cost area being fixed and then using something like this to draw a dynamic item grid. I am not sure how the owner figures out the magic number but it somehow works out.
https://imgur.com/a/dynamic-draw-itemgrid-buttongrid-V3NYOuB

Hopefully this essay is not too confusing to read with so much English, sorry my only language...

But this is an interesting refactor implementation to tackle, so if I have time I might also try to work on it.  If you beat me that's fine.  I don't have as much time anymore with real life starting to pick up again. I stood up general shop a long time ago so don't really remember a lot but if I focus can probably remember a little bit again haha.

nEEtdo0d:
ongoing drafts, not finished but the idea is there
https://pastebin.com/LsC7gUxv
https://imgur.com/a/mild-progress-4iDUtzs
shows capable able to match templates and button areas

refactored shop routine draft
https://pastebin.com/qAA8N6j0

will try getting back to it just not sure how quick
mock up refactor plan
https://pastebin.com/EvWEmr56

nEEtdo0d:
https://imgur.com/a/item-grid-almost-there-vXTXjOO

@ChthoIly @dragonheart107 pretty sure no one would have a sample image so maybe have to wait to next week, need the item page with the roses _**not**_ bought to make sure the grid I have set up is still able to detect coin and price of roses

server should not matter more or less, should be able run through a tester to gauge

this helps detect the sold out status since both are gone hence why in the image it doesn't say roses is among the choices

please and thanks

ChthoIly:
> pretty sure no one would have a sample image

Good news to you, I do have the sample image. I purposely left the rose and cake for testing and dev. So, I still got them. Here is a original 1280x720 resolution image of the `/store/sirius/gift` page:
MuMu-20250717-201703-189

From what I observed, the order of the item list changes depending whether the items are sold out or not, therefore a scan of the list must be done and uses of fix position assumption algorithm will not work. It must handle the page like `shops` or `commission` where it can scan the area/list and see where the target item is.

BTW, how is you progress? I am still working on image recognition.

nEEtdo0d:
| From what I observed, the order of the item list changes depending whether the items are sold out or not

After the purchase is made while still in the same screen without exiting, know if the item moves immediately after so shifting the `cake` as the 1st in your sample?

| Therefore a scan of the list must be done and uses of fix position assumption algorithm will not work

Yeah unfortunately as can be seen across our combined images

| BTW, how is you progress? I am still working on image recognition.

I think I have a pretty good grid for image recognition that is pretty consistent across both your sample image and mine, it detects the item pretty well.  A problem I ran into is the scale of the coin and gem are different so in the run it often crops this too much.  The trick is to use an even smaller template and not the whole icon, basically restrict to just the inside part of the icon.

But otherwise, I get the price so this is a way to detect the sold out status without relying on clicking it like I originally planned.

I am moving towards the page switching, the sirius one and the all vs gift page.

nEEtdo0d:
@ChthoIly 
Shows new entering shop routine, scans the items, and leaves
https://github.com/user-attachments/assets/0d4192a1-c5b2-450b-8651-334884568e2a
https://pastebin.com/tSgzyDtM

Shows the same but buys cake which I didn't let it lol, just to show it is able to pick the item too and do buy attempt
https://github.com/user-attachments/assets/070a4a20-8603-4780-a4f7-a1aa144161e4
https://pastebin.com/hUgvEy5E

refactor branch
https://github.com/LmeSzinc/AzurLaneAutoScript/compare/master...nEEtdo0d:AzurLaneAutoScript:private_quarters_refactor

This should hopefully address the bug altogether plus add support to read gems and buy cake if enough, like roses bought in bulk or none at all

However to complete the refactor I will need an example image of what it shows after pressing confirm on cake.  It should be something like the roses counterpart, the one that is in the center.  So if you can provide the image or if you already have cleaned asset to use, I can use that too.

If there is overlap icon or asset that can be used in both, that would be nice but I already forgot what roses look like.  Not sure if you have a sample image of the roses one since I do not anymore.

Have to wait to next week to strictly test to make sure the new buy routine works in full before confident to post a PR

Thanks again for the report and encouraging this, did some things I wanted to do for the original but got lazy haha

ChthoIly:
@nEEtdo0d 
Just saw your message. Did you finished your code?

I just wrapped up a minute ago. I got away with Template matching, when I came across reading the [Template](https://github.com/LmeSzinc/AzurLaneAutoScript/wiki/4.1.-Detection-objects#template) docs. Used your logic on checking for the price bar to see whether the item is available or sold out.

Seems to work stable for me. Tested on both `Cake` and `Rose` worked well on my CN device. Was able to locate and buy as well as understand is sold out.

Here is a short video sample on first run trying to buy the cake and second run to test for sold out check. Pass the test well.

https://github.com/user-attachments/assets/e53050b2-6091-4f49-ae61-24ad0253f0cd


https://github.com/user-attachments/assets/db4f9871-0194-4b31-b7c5-6c2369af01d8

module/private_quarters/private_quarters.py
```diff
@@ -216,84 +262,153 @@ def pq_shop_weekly_roses(self):
...
+        Execute purchase operation for a single item
+        item_key: 'rose' or 'cake'
+        """
+        item = self.shop_weekly_items[item_key]
+        logger.hr(f'Try to buy {item["name"]}', level=2)
+
+        # Identify if the item exists and get its position
+        self.device.screenshot()
+
+        item_box = self.find_button_box(item['item_img'], similarity=0.85)
```
ChthoIly, line 362: Would be great if someone can test this on EN/JP server. Hopefully localization difference won't impact template match similarity too much.

nEEtdo0d:
> I just wrapped up a minute ago. 

Lol I just finished my version too.  Seems we both do the job pretty well.  I also took your `WEEKLY_CAKE_GET` which is what I needed to complete my branch altogether, so thank you for that!  Still like to give it a full go but it will have to be an item that I'm willing to buy i.e. roses.  I also plan to change the interface for enabling roses and cake to string just like shop for consistency since now its more than 1 item.

> I got away with Template matching, when I came across reading the [Template](https://github.com/LmeSzinc/AzurLaneAutoScript/wiki/4.1.-Detection-objects#template) docs.

Yes this is the only way to do that especially if the icon you are looking for can be anywhere on the screen instead of a relative fixed position.  My version also uses template matching, I just use the helper routines provided by other modules so little less duplicate and more reuse.

> Would be great if someone can test this on EN/JP server. Hopefully localization difference won't impact template match similarity too much.

Well the templates you have include CN text in them.  So do not think it will work.  If you capture instead just the item like my version does, it might be better localization.  My version also caters to localization, so would suggest cloning or configuring ALAS config to https://github.com/LmeSzinc/AzurLaneAutoScript/compare/master...nEEtdo0d:AzurLaneAutoScript:private_quarters_refactor

To see whether runs as expected on CN, @guoh064 for JP

If it works, I think should go with my version as more reuse and encapsulation like `Navbar, ShopBase, ShopClerk, and etc.` also a little more organized rather than all in 1 file, didn't like that for awhile but seeing as it did not have many function definitions it felt fine but this refactor grew it so split across files and looks more like the `shop` module

@ChthoIly in your case since you did buy everything.  When attempting to buy, it will scan the items and find nothing because nothing has price and cost to them shown hence detected as sold out naturally rather than depending on an asset. If it does this, then I think should work out for you next week when they do have prices on them.

module/config/i18n/en-US.json
```diff
@@ -2060,6 +2060,10 @@
       "name": "Buy Weekly Roses",
       "help": "On week reset, roses will be available for purchase\nThe purchase is made in bulk wholsesale or none at all\nMinimum of 24K coin is required, if not enough subtask will be delayed to the next day and try again then"
     },
+    "BuyCake": {
+      "name": "Buy Weekly Cake Gift",
+      "help": "Cake gift purchase count is reset every week\nThe purchase is made in bulk wholsesale or none at all\nMinimum of 150 gems is required, if not enough subtask will be delayed to the next day and try again then"
```
nEEtdo0d, line 2065: This is actually 210

Total of 5 can be bought and price at 30 but the price changes as you increase

This was also the case for roses, initially priced 1000 but actually each bump up increased the price even more.  Pretty naughty of them...

nEEtdo0d:
@ChthoIly my version is in final draft, the `BuyRoses` toggle has been removed replaced with `Filter` instead and just like shop, use filter string for your case since you want to buy cake `GiftRoses > GiftCake` for me and F2P its just `GiftRoses` haha

If alright, can provide text for this in CN for this config and I will push into my version. Thanks!

image vs 
image

A link for convenience https://github.com/LmeSzinc/AzurLaneAutoScript/commit/678035b0a1e05a8b404754754a4bad4a06348441#diff-fbf7415162025c16291fa77bc550b61519e87c1ec673b34c755b8bcf1305245fR2061

ChthoIly:
> If alright, can provide text for this in CN for this config and I will push into my version. Thanks!

Just added comment for CN localization on your version, here is also a copy of it if you need it.
`"help": "仅限两种物品，浪漫满分和蛋糕礼盒\n每周一重置购买\n购买为一次性全部购买\n购买浪漫满分需要24000物资\n购买蛋糕礼盒需要210钻石\n如果任一物品的资金不足，该子任务将延迟到次日重试\n若要禁用此功能，在过滤器中输入空格字符"`

Can't wait to see it update to the master branch 👍👍👍

nEEtdo0d:
> Just added comment for CN localization on your version, here is also a copy of it if you need it. `"help": "仅限两种物品，浪漫满分和蛋糕礼盒\n每周一重置购买\n购买为一次性全部购买\n购买浪漫满分需要24000物资\n购买蛋糕礼盒需要210钻石\n如果任一物品的资金不足，该子任务将延迟到次日重试\n若要禁用此功能，在过滤器中输入空格字符"`
> 
> Can't wait to see it update to the master branch 👍👍👍

Thank you for the text, pushed and posted draft PR <https://github.com/LmeSzinc/AzurLaneAutoScript/pull/4967>.  Consider closing this one.  Please verify if localization is stable either through configuring ALAS config or cloning the branch and running independently

Thanks again and sorry for preference my version over yours.

PR #4964, opened 2025-07-16
Fix: [CN] PRIVATE_QUARTERS_INTERACT button

sui-feng-cb:
修复#4961，直接复制了en的资源，可能是PRIVATE_QUARTERS_INTERACT和PRIVATE_QUARTERS_INTERACT.BUTTON两张图片中的区域高度不一致，导致offset设置后位置不正确
 
 另外，#4942 好像把#4945 里改过的日服的ui/assets.py里的PRIVATE_QUARTERS_CHECK改回来了，在运行button_extract的时候一起修改了


nEEtdo0d:
Before release this also happened after adding support for new jersey but after applying this, it became more consistent.  https://github.com/LmeSzinc/AzurLaneAutoScript/blob/master/module/private_quarters/private_quarters.py#L278

I figured because new jersey has just 3 options whereas others have 4.  The EN asset was captured using anchorage as the sample so odd.

sui-feng-cb:
> Before release this also happened after adding support for new jersey but after applying this, it became more consistent. https://github.com/LmeSzinc/AzurLaneAutoScript/blob/master/module/private_quarters/private_quarters.py#L278
> 
> I figured because new jersey has just 3 options whereas others have 4. The EN asset was captured using anchorage as the sample so odd.

Yep, the number of options for Anchorage and New Jersey is different.

For CN server, [PRIVATE_QUARTERS_INTERACT](https://github.com/LmeSzinc/AzurLaneAutoScript/blob/master/assets/cn/private_quarters/PRIVATE_QUARTERS_INTERACT.png) was captured using Anchorage, which is the same as EN asset. But for [PRIVATE_QUARTERS_INTERACT.BUTTON](https://github.com/LmeSzinc/AzurLaneAutoScript/blob/master/assets/cn/private_quarters/PRIVATE_QUARTERS_INTERACT.BUTTON.png), they used New Jersey.

That causes `PRIVATE_QUARTERS_INTERACT` having a different detection area and click area. 
(https://github.com/LmeSzinc/AzurLaneAutoScript/blob/master/module/private_quarters/assets.py#L8)
Its y-coordinates of the click area is lower than in the EN asset. But in other servers, they are the same. After applying `_button_offset`, the click position is still below the expected position, causing accidental clicks on the gift button. I think CN assets should either use both Anchorage or New Jersey.

nEEtdo0d:
Agreed besides anything with `*.BUTTON` is not what is used to match against a screen capture so its okay to stick with the EN text version.  The mismatch might be throwing off ALAS because its adding the offset as to where it detects the icon and likely adds to the existing coordinates of the `.BUTTON` hence overlapping with the gift select option probably.

So long as the area and position for this restored asset still lines up in CN then I think should be fine.

If icon or shape differs then better to start from scratch and cut out new pair assets altogether.

PR #4967, opened 2025-07-19
Opt: Private Quarters refactor

nEEtdo0d:
- Addresses stale roses sold out reported bug (#4943), instead of purchase through front page, moves to a different page in shop.  This allows game to properly re-query data to load an accurate representation of what actually is available in private quarters shop
- Expand purchase mechanism to buy both roses (coins) and cake (gems), use filter mechanism to customize

Thank you @ChthoIly for reporting and discuss (sorry we implement together but hopefully learned about ALAS)
Thanks @dragonheart107 and @guoh064 for participating

TODO:
- @guoh064 found due to lack of info bar, may also address how to better detect page arrival or alternative strategy.  [Early screen capture results in this](https://imgur.com/a/jp-page-appear-may-apply-to-other-servers-too-TpAQmDz)

In draft until can confirm all operations were unhindered and operate appropriately for participating servers


ChthoIly:
Noticed a small issue in the Gift Purchasing process.

Seems missing a confirm action after purchase which made the program to through a timeout exception. Just simple a click action after successful purchase should resolve the issue.

Screenshot 2025-07-20 170930

Here is error log file if you need it.
[ErrorLog-MissingConfirm.zip](https://github.com/user-attachments/files/21335922/ErrorLog-MissingConfirm.zip)

Apart from that seems to work fine in CN server with daily interaction and gift identification and sold out identification.

nEEtdo0d:
> Seems missing a confirm action after purchase which made the program to through a timeout exception

Oh no, that part did not change though, so that is odd.  Thank you for the images!

Both the cake and roses images overlay perfectly 1 to 1 between screen capture and asset, so something must have went wrong with the if condition...

nEEtdo0d:
> Seems missing a confirm action after purchase which made the program to through a timeout exception

Fixed it.  Remembered I did change something about that area.  The sample images helped so thank you, I didn't realize the shop check button was blurry I thought it was just overlaid with transparent black so technically could still be seen and evaluate to true.

Although it isn't a live test, I tried to simulate it best I could.  I fed in your image first then had device screen capture after, it sort of simulated the scenario and I'm pretty confident it should run appropriately now.  Sorry about that!!

ChthoIly:
Good to know.

Is there anything else waiting to be tested or implemented before pushing to the master branch, or just waiting for LmeSzinc to review it?

nEEtdo0d:
> Is there anything else waiting to be tested

Verify feature runs through both shop and interact still works after refactor for all servers.  This bug you found was huge, would have been sad if I ran this tomorrow and it didn't work out haha...so thank you for bringing it up.  I think you've more or less confirmed still good for CN, you had to work around the bug but the interact was still stable right?

Before moving out of draft, want to run the task in full with both enabled for EN.  After that, will move to ready and wait owner's review, besides the shop portion, everything else is still the same so hoping having been okay with the first version, shouldn't have too many questions about it.  Also advertises with reuse instead of reinvesting the wheel.

> or implemented

There is but it might not make it until we get an idea how to address it, so it won't block if this PR is totally ready without it.  Those with higher end emulators enter the private quarters too fast, gets a bad image but a good enough image where it evaluates that it did enter the page.  But as a result, parts of the screen are absent like the daily count area and shop enter area.  So ALAS unintentionally skips over.  As low end emulator user, I do not come across this problem.

ChthoIly:
> the interact was still stable right?

Yes, interact always went smoothly on my device. Except the purchase checkout issue just mentioned all `private_quarters` components worked like a charm to me.



> Those with higher end emulators enter the private quarters too fast, gets a bad image but a good enough image where it evaluates that it did enter the page.

I do see there are quite a few issued opened recently with private quarters, mainly on interacting and navigation. However, I can't reproduce any of them on my device. Not sure if the push could solve them. 

I am using MuMu 12 with a 98X3D and 3090 and everything on the emulator cranked to the maximum setting. Didn't notice any navigation or screenshot issues. The report on these issues are not so clear to me, and I am having a hard time trying to reproduce them. Is there any clear detail about the screenshot issue? like what devices are used and what to be considered as 'high end'?

nEEtdo0d:
> I do see there are quite a few issued opened recently with private quarters, mainly on interacting and navigation. However, I can't reproduce any of them on my device. Not sure if the push could solve them. 

Yikes, didn't think to check issue.  Oh jeez all in CN too, I'll respond but we'll see whether my message gets across haha...

> The report on these issues are not so clear to me, and I am having a hard time trying to reproduce them

I gauge most of them to be either setting issues so ALAS is using the wrong assets like using EN server assets on CN server for example.

Emulator of choice can also be factor in performance in the swipe emulation.  If I had to guess, the swipe isn't triggering and so the image stays the same throughout every check. 

> Is there any clear detail about the screenshot issue? like what devices are used and what to be considered as 'high end'?

This link belongs to a JP run, this the image captured at the time of page switch.  It does not retake any new screenshots and uses this to do the OCR, there is nothing to OCR on it and as result evaluates to 0 and exits the run.  https://imgur.com/a/jp-page-appear-may-apply-to-other-servers-too-TpAQmDz

If we can use `skip_first_screenshot=False` more liberally, it would fix most issues however it is expensive performance wise to rely on that so often is rejected to do this to resolve problems.

The reporter did not mention specs or device information.  But basically their emulator is beefy enough that this line in code doesn't execute because no info bar on screen, `self.handle_info_bar`.  This ensures more images are taken due to slow emulator and increases the chance of the final image being fully loaded page.

nEEtdo0d:
@ChthoIly a interesting behavior after purchase, it will reset back to the all tab page and next scan will not pick up cake since its off screen again so inserted recover steps back so can smoothly make both purchases in one task run

Before moving out of draft.  To make sure I don't feel like I'm missing anything else since I got the chance to run and ran into this, I will wait for few hours to think on it.

ChthoIly:
For me the last I run the task, I set `cake > rose`. So I ran `cake` first, and since it failed to confirm purchase and timed out. I manually started the task again for rose. Therefore, did not notice this issue. Worth to notice.

module/private_quarters/interact.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Args:
+            target_ship (str):
+
+        Returns:
+            bool
+        """
+        target_title = target_ship.title().replace('_', ' ')
+        page_btn = self.available_targets[target_ship][1]
+        logger.hr(f'Seek {target_title}\'s Page', level=2)
```
LmeSzinc, line 64: Possible KeyError and IndexError in `self.available_targets[target_ship][1]`, log and skip if input ship is invalid

module/config/argument/argument.yaml
```diff
@@ -586,7 +586,8 @@ SupplyPack:
 Minigame:
   Collect: false
 PrivateQuarters:
-    BuyRoses: true
+    Filter: |-
+      GiftRoses
```
LmeSzinc, line 590: Maybe just keep `BuyRoses: bool` for easier understanding. I think there won't be more gifts in the future, no need to have such high customization and extensibility.

module/private_quarters/clerk.py
```diff
@@ -0,0 +1,100 @@
...
+                )
+                break
+
+            if self.appear(PRIVATE_QUARTERS_SHOP_CHECK, interval=3):
+                self.device.click(item)
+                continue
+            if self.appear_then_click(PRIVATE_QUARTERS_SHOP_AMOUNT_MAX, offset=(20, 20), interval=1):
+                continue
+            #if self.appear_then_click(PRIVATE_QUARTERS_SHOP_CONFIRM_AMOUNT, offset=(20, 20), interval=1):
+            #    continue
```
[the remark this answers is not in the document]
nEEtdo0d: The `ui_click` was to address report of a click but did not get acknowledged by game so stayed in the screen then got stuck in `_pq_shop_exit()` as a result.

But splitting it should not be too much trouble

PR #4998, opened 2025-08-02
Fix: dropscreen with no hardcoded sleep (#4907)

dragonheart107:
replaced the hardcoded sleep with an area change detection, it should work for CN, but might have issues in TW or JP since their base asset used as anchor point is non default for JP and smaller for TW  (#4907)
kept debug related parts in for now

additionally 2 asset related fixes for EN, non urgent -> added in this PR:
dock amount: position and size changed, Ocr misdeteced
enhance: alas would never flag it as empty:
```
        def state_enhance_recommend():
            # Judge if enhance material appeared
            if not EMPTY_ENHANCE_SLOT_PLUS.match(self.device.image):
                logger.info('Material found. Try enhancing...')
                return "state_enhance_attempt"
            elif self.info_bar_count():
                logger.info('No material found for enhancement.')
                logger.info(
                    'Enhancement failed. Swiping to next ship if feasible')
                return "state_enhance_fail"

            return "state_enhance_ready"
            ```


LmeSzinc (commented):
1. `CampaignBonusStatistics` should not inherit `ButtonGrid` and `Timer`. It will pollute the entire Alas module
2. Don't use ocr if you don't need the exact number
3. Never use `cv2.imread` and `cv2.imwrite`. opencv can't handle non-ascii filepath, and so many years passed by they never fixed it.

LmeSzinc:
One PR do one thing. Open another PR for the DOCK_AMOUNT thingy

dragonheart107:
> 1. `CampaignBonusStatistics` should not inherit `ButtonGrid` and `Timer`. It will pollute the entire Alas module
> 
>     2. Don't use ocr if you don't need the exact number
> 
>     3. Never use `cv2.imread` and `cv2.imwrite`. opencv can't handle non-ascii filepath, and so many years passed by they never fixed it.

~~for 1, create it's own class then?~~
2. well I kinda needed a way to verify grid and ocr was kinda the easiest since every drop has a number/amount; do you have alternative suggestion?
~~3. this shouldn't be an issue since these are only debug related no? would be commented or removed afterwards after getting the ok from you for the rest?~~

dragonheart107:
I don't really have an idea how to implement it without the OCR, since it is way to dynamic for any other check. If you have an idea pls share @LmeSzinc.

Aside from that, adjusted to your other requests;
also thinking about adding a scrolling part in since some reward screens have so many rewards that you need to scroll to see them all

LmeSzinc:
Maybe post the debug images you collected, I'll try to find some features.

dragonheart107:
I don't have any debug images for the OCR part anymore after finetuning it, but the debugs I have are like this:
these are the ones from `reward_area_stable_check`
20250804_053245_0_initial
20250804_053245_1_check
20250804_053245_2_check
20250804_053245_3_check
20250804_053245_4_check
20250804_053245_5_check
20250804_053245_6_check
20250804_053245_7_check
20250804_053245_8_check
20250804_053245_9_check
20250804_053245_10_check
20250804_053245_11_check
20250804_053245_12_check
20250804_053245_13_check
20250804_053245_14_check
20250804_053245_15_check
20250804_053245_16_check
20250804_053245_17_check
20250804_053245_18_check
20250804_053245_19_check
20250804_053245_20_check
20250804_053245_21_check
20250804_053245_22_check
20250804_053245_final_stable
from `wait_until_reward_stable` (this kinda unneeded now and doesn't do much, used it for a very early version to check px -relation based on exit buttons)
20250804_053245_reference_points
with this being the dropscreen:



Initially I thought just a fixed area of the reward area would be fine, but then I had issues with reward screens when they had meta xp too, moving the UI sideways; so after testing I arrived at the OCR solution
`CAMPAIGN_BONUS` asset is unreliabel for en at least as it can be Clearing Reward/Total Reward or Reward

LmeSzinc:
Crop REWARD_AREA and TITLE_REWARD, TITLE_CLEARING_REWARD using the same image, use a large search area to find the title `self.appear(TITLE_REWARD, offset=(200, 20))`, if it appears, do `REWARD_AREA.load_offset(TITLE_REWARD)`, `REWARD_AREA.button` will be moved accordingly.

dragonheart107 (commented):
had no issues testing the refactor, so should be good to merge after your OK

dragonheart107:
@LmeSzinc could you see if it's okay to merge now?

PR #5032, opened 2025-08-15
new Shop related

dragonheart107:
just a couple of new EN shop assets, might need different ones after shop logic has been adapted etc...


dragonheart107:
hm having trouble with getting too many actives from the navbar in these
MuMu-20250815-234001-751
tried adjusting active_threshold and active_count /inactive_threshold and inactive_count from the Navbar, but not really getting satisfying results

nEEtdo0d:
> tried adjusting active_threshold and active_count /inactive_threshold and inactive_count from the Navbar, but not really getting satisfying results

Haven't started myself on this, but what colors are you using as active and inactive? I was thinking maybe using the text pixel colors as those since they seem pretty distinct on and off wise.

Also hoping that it be more transparent this way for others servers, less dependency on them capturing specific server assets

The grid that navbar uses would be text centered instead of like specific section.  I had to do that for private quarters, couldn't rely on text.

nEEtdo0d:
also thanks for the item grid numbers, will help when I do tackle this but it looks like another ongoing PR is quite a bit further along so may not get there fast enough haha

still will try but it kind of had the same basic idea I was going to approach it as too.

dragonheart107:
> also thanks for the item grid numbers, will help when I do tackle this but it looks like another ongoing PR is quite a bit further along so may not get there fast enough haha
> 
> still will try but it kind of had the same basic idea I was going to approach it as too.

ye, that approach seems overall cleaner and less messy, will just keep this draft alive if we need grid coordinats/templates etc.

Edit: my implementation kinda works, but for some reason the item template matching for medal only is screwed. it somehow never matches and wants to create new templates all time (and scroll is another can of worms), other shops work fine, item recognition is fine too except some weird hick-ups for books in general store and plategeneral and plateplane being recognized as each other in one of the other shops

PR #5065, opened 2025-08-23
Upd: GUI text

dragonheart107:


nEEtdo0d:
if no activity after I address my open PR, can merge since this is just updating GUI text, no functional updates to it

PR #5073, opened 2025-08-25
Fix: voucher shop should use old ui

sui-feng-cb:
大世界白票商店的ui没有更新


nEEtdo0d:
addressing in this bulk PR in addition there were some parts also missed with this fix just mainly the `shop_interval_clear` that also needs to be overridden due to `BACK_ARROW` changing in base class, check https://github.com/LmeSzinc/AzurLaneAutoScript/pull/5078

if acceptable, can close this without merge

PR #5074, opened 2025-08-26
Fix: EN supply pack + shop timeout

dragonheart107:
self.ui_goto_shop() in shop_reward.py would lead to timeout on EN


nEEtdo0d:
got couple of hot fixes I wanted to put in as well, not everything since I went a different direction in terms of arrangement, my fork https://github.com/LmeSzinc/AzurLaneAutoScript/commit/7985542e8c4bd5412fd42f2c5a39301acec5f507 if want to gloss it 

but stuff like this shop_status because voucher shop did not change in interface and its ocr is still in the same place
https://github.com/nEEtdo0d/AzurLaneAutoScript/blob/ui_refactor_shop/module/shop/shop_status.py#L24

also commented in prior PR, ocr price numbers slightly cut off just like the author showed via images we are pixels off for whatever reason, these set of numbers worked out well for general shop.  If applicable other shops as well.
https://github.com/nEEtdo0d/AzurLaneAutoScript/blob/ui_refactor_shop/module/shop_white/shop_general.py#L18
there's also the fact we won't know if devs will re-release a more uniform grid so we aren't pixels off since our side bar differs too 

oh speaking of, do the nav bars work? because when I tested with his raw numbers, they didn't so, can look over my fork's for reference
https://github.com/nEEtdo0d/AzurLaneAutoScript/blob/ui_refactor_shop/module/shop_white/ui.py#L9-L72

not related to shop but this will help expedite tactical class because will skip meta ships instead of continuously backing out every time comes across 1
https://github.com/nEEtdo0d/AzurLaneAutoScript/blob/ui_refactor_shop/module/tactical/tactical_class.py#L631-L632

dragonheart107:
> got couple of hot fixes I wanted to put in as well, not everything since I went a different direction in terms of arrangement, my fork [7985542](https://github.com/LmeSzinc/AzurLaneAutoScript/commit/7985542e8c4bd5412fd42f2c5a39301acec5f507) if want to gloss it

yeah saw that
> but stuff like this shop_status because voucher shop did not change in interface and its ocr is still in the same place https://github.com/nEEtdo0d/AzurLaneAutoScript/blob/ui_refactor_shop/module/shop/shop_status.py#L24

ye there's a PR for that out already as well (#5073)
> also commented in prior PR, ocr price numbers slightly cut off just like the author showed via images we are pixels off for whatever reason, these set of numbers worked out well for general shop. If applicable other shops as well. https://github.com/nEEtdo0d/AzurLaneAutoScript/blob/ui_refactor_shop/module/shop_white/shop_general.py#L18 there's also the fact we won't know if devs will re-release a more uniform grid so we aren't pixels off since our side bar differs too

yup saw that too, we'll see if we get it potentially this week
> oh speaking of, do the nav bars work? because when I tested with his raw numbers, they didn't so, can look over my fork's for reference https://github.com/nEEtdo0d/AzurLaneAutoScript/blob/ui_refactor_shop/module/shop_white/ui.py#L9-L72

will look over that again, wasn't able to get it to work before myself, but I used a bunch of different values iirc;
well navigation kinda works for all but medals, it kinda shits itself on that part, but general, merit, guild and core work fine:
https://github.com/dragonheart107/ALAS/blob/bf787c0ab49e947a7f29be95fcbd78412f25b84a/module/shop/ui.py#L55-L167
> not related to shop but this will help expedite tactical class because will skip meta ships instead of continuously backing out every time comes across 1 https://github.com/nEEtdo0d/AzurLaneAutoScript/blob/ui_refactor_shop/module/tactical/tactical_class.py#L631-L632

nice QoL, was thinking about somethign similar for supply packs with the new UI, where if the exclamation mark is not there, it finishes supply run, without going to supply page etc., something along the idea of that:

```
    def run(self):
        """
        Pages:
            in: Any page
            out: page_supply_pack, supply pack tab
        """
        if self.config.SERVER in ['cn', 'en', 'jp']:
            self.ui_ensure(page_shop)
            #sidebar of shop takes a bit to show
            self.wait_until_appear(SHOP_GOTO_SUPPLY_PACK)
            if self.appear(FREE_SUPPLY_PACK_EXCLAM):

                self.ui_ensure(page_supply_pack)

                if self.get_oil_shop() < 21000:
                    server_today = get_server_weekday()
                    target = self.config.SupplyPack_DayOfWeek
                    target_name = day_name[target]
                    if server_today >= target:
                        self.supply_pack_buy(FREE_SUPPLY_PACK)
                    else:
                        logger.info(f'Delaying free week supply pack to {target_name}')
                else:
                    logger.info('Oil > 21000, unable to buy free weekly supply pack')
            else:
                logger.info('weekly supply pack already claimed')

        elif self.config.SERVER in ['tw']:
            self.ui_ensure(page_supply_pack)

            if self.get_oil() < 21000:
                server_today = get_server_weekday()
                target = self.config.SupplyPack_DayOfWeek
                target_name = day_name[target]
                if server_today >= target:
                    self.supply_pack_buy(FREE_SUPPLY_PACK)
                else:
                    logger.info(f'Delaying free week supply pack to {target_name}')
            else:
                logger.info('Oil > 21000, unable to buy free weekly supply pack')
```

nEEtdo0d:
> well navigation kinda works for all but medals, it kinda shits itself on that part, but general, merit, guild and core work fine:

I see thats weird, maybe I didnt use most recent numbers before the merge, for me it was clicking a totally off range position for all the shops and buttons left bar (supply, monthly) and upper bar  (general, guild, and etc.), after adjusting them works now for all shops, general, guild, and etc.

I will fork a new branch to more align adapt with master and use the `@Config` decorator to accommodate EN's state of the shop UI plus all the other hot fixes

> nice QoL, was thinking about somethign similar for supply packs with the new UI, where if the exclamation mark is not there, it finishes supply run, without going to supply page etc., something along the idea of that:

I think its reasonable why not to cut down some time but a similar concept was applied for guild a while back but throughout time devs changed kept moving it around and owner opted in just visiting the page even if pointless, sure a little time wasted but not too terrible.

though tactical class has similar downtime waste but in comparison varies from the number of meta ships that any player is behind on, so having it sit through 5+ of them would be benefit to speed that up, worst case 20+

nEEtdo0d:
addressing in this bulk PR in addition there were some parts also missed with this fix, like porting over the overlap assets applicable to EN `SHOP_CLICK_SAFE_AREA` and etc. https://github.com/LmeSzinc/AzurLaneAutoScript/pull/5078

if acceptable, can close this without merge

PR #5078, opened 2025-08-27
Fix: [EN] shop and supply tasks + other hotfixes

nEEtdo0d:
not totally into the whole price area adjustment for servers, focused on reducing duplication where can.

was considering on consolidating into a single function in the same fashion as `shop_grid` something like `_shop_price_area` but because CN does have different tuples in theirs for respective shops I figured it would not make a difference whereas for EN the same tuple is fine for every variant of shop

other hotfixes include
- left bar tab for shop ui position is different on EN (might be able to use same numbers after next update though which case will readjust)
- shipyard inherited general shop for the same reason as gacha so mirrored
- reduce downtime on select ship for tactical class, filter out the meta ships
- voucher shop reverted since did not change in shop ui

medal shop scroll bar out of my expertise, sorry to say so still just have to keep that disabled


nEEtdo0d:
@dragonheart107 @sui-feng-cb of course feel free to review if you feel like I've missed something or if you give it run can tell me if anything went wrong

but doing couple of trial runs, seems pretty stable

PR #5081, opened 2025-08-28
Fix: [EN] shop task; grid now uniform with other servers

nEEtdo0d:
well almost...shop tab still unique I guess because of the english text so shift button positions slightly maybe

but there is also the weird thing about the supply pack oil position, might change again since this is just right after the update we either let this sit for a little longer in case they want to rung out the rug on me again or merge ahead and fix later again

but otherwise all stable even medal shop scroll bar, really was related to version of shop interface between EN and CN, cool beans

back to regular tuples instead of `price_area` after getting back from work

sorry this happened, thanks

probably better to just let this sit for few days in case, if there is complaint from EN, can merge as is though since it does work just trying to avoid the game devs whims


assets/en/ui/SUPPLY_PACK_CHECK.png
LmeSzinc, line 1: don't crop the entire button, there will be red dot on thet right and time limited icon on the left

nEEtdo0d: fixed, tried to match as closely in crop size to CN

again thanks for catch

PR #5173, opened 2025-09-27
Fix: Guild

dragonheart107:
#5168
potentially TODO JP + TW


nEEtdo0d:
if I recall there was a reason why it used the entire text vs just dispatch, it could not tell the difference between active vs inactive or something

also the captured assets look they might factoring in a little bit of the background animation playing?

maybe just needs a fix in the condition logic that uses the asset instead of whole asset replace, think it uses `match_template` or some specialized func rather than `appear`?

dragonheart107:
> if I recall there was a reason why it used the entire text vs just dispatch, it could not tell the difference between active vs inactive or something

hm i see, my guild is doing boss fight so can't really verify with live footage + post boss fight need to wait for new selection

> also the captured assets look they might factoring in a little bit of the background animation playing?

i think it's not an animation, but rather the transparency of the button got changed and shows the ui element from previous step

> maybe just needs a fix in the condition logic that uses the asset instead of whole asset replace, think it uses `match_template` or some specialized func rather than `appear`?

seems to just be the appear
https://github.com/LmeSzinc/AzurLaneAutoScript/blob/dbf40287395eca4a5d149f82eea55a53ff78cb4b/module/guild/operations.py#L367

LmeSzinc:
use template match on `GUILD_DISPATCH_FLEET_UNFILLED` and see if it matches.
If yes, use template match on `GUILD_DISPATCH_FLEET` and add `image_color_count()` counting white pixels for double check.

PR #5174, opened 2025-09-27
Add: PrivateQuarters Taihou option

nEEtdo0d:
Add Taihou option to PrivateQuarters task, addressed language yamls en and cn (copied from Azur Lane wiki)

Handful number of times, the check asset appeared for an instant on room enter so figured a constant interval reset will help skip over it temporarily until stable. Hasn't happened after patching it so not totally sure whether this completely fixes it but I think it will if it does.


module/private_quarters/interact.py
```diff
@@ -164,16 +165,18 @@ def _pq_goto_room_enter(self, target_ship):
                 self.device.screenshot()
 
             # End
-            if self.appear(PRIVATE_QUARTERS_ROOM_CHECK, offset=(20, 20)):
+            if self.appear(PRIVATE_QUARTERS_ROOM_CHECK, offset=(20, 20), interval=1):
                 break
```
LmeSzinc, line 169: end condition should not have interval

nEEtdo0d: I can remove the commit altogether if that is preferred 

But yeah at risk of getting stuck because of this on and off thing that might appear right before character dialogue

LmeSzinc: maybe saving continuous screenshots of page switching, I'll try to to find some features

nEEtdo0d: > maybe saving continuous screenshots of page switching

like retaining the last 10 screenshots or so then reviewing them after the load sequence is done to make sure that UI element didn't stick around long enough to cause accidental exit

> I'll try to to find some features

please do, if there are existing examples of doing something like this I can adopt it

this just seemed the easiest to do using native timer to block it but I can understand the concern too

nEEtdo0d: > can you take the screenshots?

sure
[screenshots.zip](https://github.com/user-attachments/files/22628661/screenshots.zip)

I also got some footage captured showing the asset appearing right before dialogue and without. Wasn't expecting to get it the first try because has happened in a while but there it is haha. all zip files since github limits 25MB uploads heh.
[video of lagging asset 1.zip](https://github.com/user-attachments/files/22628662/video.of.lagging.asset.1.zip)
[video of lagging asset 2.zip](https://github.com/user-attachments/files/22628663/video.of.lagging.asset.2.zip)
[video of no lag.zip](https://github.com/user-attachments/files/22628664/video.of.no.lag.zip)

nEEtdo0d:
Still sifting through options and digging ways to streamline this better, already thinking its pretty good but that damn pause

At the same time planning a vacation so may not get around to this for awhile

This isn't priority update anyway so can be held off for a long time

nEEtdo0d:
Sorry for the delay

Pushed now and does as suggested, inserts handling of dialogue at other transitional states besides room enter

Funny thing is though...after running this task 20+ straight times using a loop, the lag never appeared

So can consider not having this commit altogether since it appears the devs on their side maybe did some optimizations I guess on their end.

PR #5181, opened 2025-09-29
Fix: Return to main page and back after can't get oil amount

JerryZhousy:
在无法获取油量/油量为0时回到主界面再返回
https://github.com/LmeSzinc/AzurLaneAutoScript/issues/5163


dragonheart107:
feels a bit redundant since it was fixed on CN a few days ago + todays patch for JP/EN, not sure about TW

PR #5230, opened 2025-10-17
Fix: EN commission + mission reward

dragonheart107:
i opted to change dictionary in project data since some commission title names are still overlapping onto the chibi and this worked

unsure wether the ui page appear instead of ..._check was only due to en or if you wanted to keep it


dragonheart107:
MuMu-20251017-173000-855

module/commission/commission.py
```diff
@@ -518,7 +518,7 @@ def commission_receive(self, skip_first_screenshot=True):
                     self.device.screenshot()
 
                 # End
-                if self.ui_page_appear(page_commission, offset=(20, 20)):
+                if self.appear(COMMISSION_CHECK, offset=(20, 20)):
```
LmeSzinc: keep these two changes

module/reward/reward.py
```diff
@@ -118,7 +118,7 @@ def _reward_mission_collect(self, interval=1, skip_first_screenshot=True):
                     timeout.reset()
                     continue
 
-            if not self.ui_page_appear(page_mission):
+            if not self.appear(MISSION_CHECK):
```
LmeSzinc: 2

LmeSzinc:
Noticed that there is a submodule added and removed, `Submodule AzurLaneAutoScript added at 9b0e51`. I have to squash the entire PR to avoid adding it to commit history.

dragonheart107:
> Noticed that there is a submodule added and removed, `Submodule AzurLaneAutoScript added at 9b0e51`. I have to squash the entire PR to avoid adding it to commit history.

ye I was confused myself, not quite sure how it ended up here either

PR #5237, opened 2025-10-18
Upd: adjustments for EN commission titles

dragonheart107:
had some time to test and chibis on commission didn't impact the title ocr


dragonheart107:
let it run the past days and seemed more stable than current version

Issue #5692, opened 2026-06-01
[Bug]Daily 6 fail: Too many click for a button: DAILY_ENTER


xlgamer2:
### Before you asking

- [x] I have searched the existing issues
- [x] I spend at least 5 minutes for thinking and preparing
- [x] I checked Frequently Asked Questions (FAQ) in Wiki page
- [x] I am using the latest version of Alas

### Describe the bug

Alas would get stuck at Daily quest's 6th tasks here. It doesn't happen every time, just sometimes.

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/196b527b-19cf-4389-b774-154ccf9d9ae7" />

### To Reproduce

1. let alas run daily quest
2. get stuck at daily 6

### Relevant log output

```Text
─────────────────────────────────────────────────────── DAILY 6 ───────────────────────────────────────────────────────
2026-06-01 02:14:47.657 | INFO | DAILY 6                                                                               
2026-06-01 02:14:47.658 | INFO | remain=3, stage=1, fleet=1                                                            
2026-06-01 02:14:47.659 | INFO | <<< UI CLICK >>>                                                                      
2026-06-01 02:14:47.662 | INFO | Click ( 665,  403) @ DAILY_ENTER                                                      
2026-06-01 02:14:57.895 | INFO | Click ( 629,  445) @ DAILY_ENTER                                                      
2026-06-01 02:15:08.148 | INFO | Click ( 661,  567) @ DAILY_ENTER                                                      
2026-06-01 02:15:18.397 | INFO | Click ( 653,  353) @ DAILY_ENTER                                                      
2026-06-01 02:15:28.654 | INFO | Click ( 671,  391) @ DAILY_ENTER                                                      
2026-06-01 02:15:38.904 | INFO | Click ( 652,  477) @ DAILY_ENTER                                                      
2026-06-01 02:15:49.161 | INFO | Click ( 652,  352) @ DAILY_ENTER                                                      
2026-06-01 02:15:59.411 | INFO | Click ( 636,  359) @ DAILY_ENTER                                                      
2026-06-01 02:16:09.665 | INFO | Click ( 701,  310) @ DAILY_ENTER                                                      
2026-06-01 02:16:19.918 | INFO | Click ( 672,  405) @ DAILY_ENTER                                                      
2026-06-01 02:16:30.170 | INFO | Click ( 680,  464) @ DAILY_ENTER                                                      
[...]

PR #5694, opened 2026-06-03
Remove duplicated button assets, and fixed outdated DAILY assets.

Tianming-Wu:
I found that although the button_extractor fallbacks to cn when it did not find the correspond asset, there was still many duplicated assets in en, jp and tw. I wrote a script to delete duplicated assets.

Also, DAILY_LOCKED and DAILY_ACTIVE is outdated, and can probably cause problems sometimes #5692 .

This is how I try to fix this.

I haven't got a chance to test it since Alas has already completed daily task on my account. I will try to test it tomorrow, if you want it to be tested before being used.

我为什么要发英文，github上说英文说习惯了（


LmeSzinc:
update daily assets only, make another PR to remove duplicate assets

PR #5696, opened 2026-06-04
Daily fix

Tianming-Wu:
DAILY_LOCKED and DAILY_ACTIVE is outdated, and can probably cause problems sometimes #5692 .

This is how I try to fix this.

I haven't got a chance to test it since Alas has already completed daily task on my account. I will test it tomorrow, if you want it to be tested before being used.


Tianming-Wu:
image

Old recognizing zone now contains red text, and since there are particles, it might get just over the threshold under very specific case. It might not occur in cn, but may occur in en. I haven't got any clue from jp or tw.

Tianming-Wu:
Tested locally, everything's fine.

I can't test en, tw and jp though, but I checked the UI and they are the same in the selected region.

PR #5700, opened 2026-06-05
Fix: Handle grayscale image matching with RGB templates matching

ycnliu:
Fixes the OpenCV matchTemplate assertion error when matching RGB template frames against grayscale images.

  ## Issue
  Campaign OCR fails with error:
  cv2.error: (-215:Assertion failed) (depth == CV_8U || depth == CV_32F) && type == _templ.type() && _img.dims() <= 2

  This occurs when `TEMPLATE_STAGE_CLEAR_20240725` (GIF with RGB frames) is matched against grayscale stage images in
  `campaign_extract_name_image()`.

  ## Changes
  - Added automatic grayscale conversion of templates when input image is grayscale in `Template.match_multi()`
  - Handles both GIF and non-GIF templates
  - Prevents type mismatch between RGB templates and grayscale input images

  ## Testing
  The fix ensures that when a grayscale image is passed to `match_multi()`, any RGB templates are automatically converted to grayscale
  before matching.


LmeSzinc:
Convert TEMPLATE_STAGE_CLEAR_20240725 to grayscale image instead. Converting channels internally may have bad performance and hard to notice

ycnliu:
This fix is superseded in the new pr.

PR #5705, opened 2026-06-06
Fix: Auto-convert pseudo-grayscale GIF templates to true grayscale

ycnliu:
## Summary
  Fixes OpenCV template matching error when matching RGB templates against grayscale images.

  ## Problem
  The `TEMPLATE_STAGE_CLEAR_20240725` GIF files are loaded as RGB arrays by `imageio` (even though they contain grayscale data with all
  channels identical). This causes `cv2.matchTemplate()` to fail with:

  error: (-215:Assertion failed) (depth == CV_8U || depth == CV_32F) && type == _templ.type() && _img.dims() <= 2 in function
  'matchTemplate'

  ## Solution
  Added automatic detection and conversion logic in `Template.image` property:
  - Detects when an RGB template has all channels equal (pseudo-grayscale)
  - Automatically converts to true 2D grayscale array during loading
  - More robust than manually converting GIF files since imageio always loads GIFs as RGB

  ## Changes
  - `module/base/template.py`: Added 7 lines of auto-conversion logic
  - Updated GIF template files during conversion attempts

  ## Test plan
  - [x] Verified template loading converts pseudo-grayscale to true grayscale
  - [x] Template shape changes from (20, 30, 3) to (20, 30)
  - [x] Application tested to confirm error is resolved and campaign is now running


guoh064:
wow AI committer

PR #5821, opened 2026-07-14
Fix MuMu Pro emulator-* serial detection on macOS

954708775:
## Summary

MuMu Pro on macOS can expose itself as an `emulator-*` serial while still advertising MuMu-specific `nemud.*` properties. The current detection logic treats every `emulator-*` device as LDPlayer/BlueStacks family, so MuMu-specific handling is skipped.

On the affected real environment, that broke the screenshot pipeline in two visible ways:

- `DroidCast_raw` screenshots were rotated by 90 degrees
- downstream page detection could fall into `Unknown ui page`

This patch adds a narrow macOS fallback: when the serial is `emulator-*` and `getprop nemud.player_engine` returns `MACPRO`, treat the device as MuMu family instead of LDPlayer/BlueStacks family.

## Why this approach

- keeps the existing serial heuristics intact for every other platform
- avoids broadening all macOS `emulator-*` devices into MuMu
- uses MuMu-specific Android properties that are already present on the real device

## Verification

Verified on a real macOS + MuMu Pro environment:

- serial: `emulator-5554`
- `getprop nemud.player_engine`: `MACPRO`
- `is_mumu_family` changed from false to true
- `is_ldplayer_bluestacks_family` changed from true to false
- `is_mumu_over_version_356` became true
- live `DroidCast_raw` screenshot returned upright after the change

## Related

Related to #5457.


module/device/connection.py
```diff
@@ -331,6 +331,40 @@ def is_bluestacks_air(self):
...
+
+    @cached_property
+    def is_ldplayer_bluestacks_family(self):
+        if self.is_mumu_family:
+            return False
+        return ConnectionAttr.is_ldplayer_bluestacks_family.__get__(self, Connection)
+
     @cached_property
     @retry
     def is_mumu_pro(self):
```
LmeSzinc, line 336: 这里已经有这个缓存属性了，不要另起炉灶

module/device/connection.py
```diff
@@ -331,6 +331,40 @@ def is_bluestacks_air(self):
...
+        logger.attr('nemud.player_engine', engine)
+        if 'MACPRO' in engine.upper():
+            logger.attr('is_mumu_emulator_serial', True)
+            return True
+
+        return False
+
+    @cached_property
+    def is_mumu_family(self):
+        if ConnectionAttr.is_mumu_family.__get__(self, Connection):
```
LmeSzinc: 别整这些 get set 魔法

954708775:
已按 review 收敛实现，感谢反馈。

我这次去掉了额外引入的 `is_mumu_emulator_serial`，也撤掉了对 `is_mumu_family` / `is_ldplayer_bluestacks_family` 的重写，不再使用 `__dict__` 手动注入缓存或 descriptor 调用。

现在改为直接扩展现有的 `is_mumu_pro` 逻辑：在 macOS 下，如果 serial 是 `emulator-*`，则通过现有的 `nemud.player_engine` 判断是否为 MuMu Pro。与此同时，把 `check_mumu_app_keep_alive` 和 `is_mumu_over_version_356` 的判断链也一并对齐到 MuMu Pro。

我理解这样更符合当前文件已有的属性结构，同时保留这次修复想解决的问题。麻烦再帮忙看下这版。

PR #5879, opened 2026-08-14
Fix: Island dock character search when owning a lot of chars

dragonheart107:


module/island_handler/dock.py
```diff
@@ -160,12 +160,16 @@ def island_dock_find_character(self, identity):
                 return candidate
             self.next_dock_page()
             if self.appear(ISLAND_DOCK_DETECT, offset=(20, 20)):
-                logger.warning('Reached end of dock page')
-                break
+                # go back to top and rescan again to get missed out chars if owning a lot of chars
+                logger.warning(f'Reached end of dock page, going back to top for one more scan for {identity}')
+                self.ensure_dock_page_at_top()
+                ISLAND_DOCK_DETECT.load_color(self.device.image)
+                ISLAND_DOCK_DETECT._match_init = True
```
guoh064: This code block seems to loop forever if a non-existing (due to either poor template or poor config) character is specified. 
I suggest searching for candidate twice in `IslandRestaurant.choose_waitress()`'s for loop and if first search gives `None` then we `ensure_dock_page_at_top` and search for once more, but not after that.

dragonheart107: the timeout of the for loop handles that no?

but sure i can have a look at modifiying `IslandRestaurant.choose_waitress()`

guoh064: Well that timeout was intended as per search timeout, so if used for two characters maybe the timeout is not enough while if you doubled the timeout then for one person that is too much.

Issue #5985, opened 2026-09-14
Debian 11 (Bullseye) end of life causing docker image build failure


fybmain:
### Before you asking

- [x] I have searched the existing issues
- [x] I spend at least 5 minutes for thinking and preparing
- [x] I checked Frequently Asked Questions (FAQ) in Wiki page
- [x] I am using the latest version of Alas

### Describe the bug

Debian 11 (Bullseye) reached its final [End of Life (EOL)](https://www.debian.org/News/2026/20260831) for Long Term Support (LTS) on August 31, 2026. Some files on the Debian official server are missing now. Need to upgrade to a new Debian version to complete the installation.

### To Reproduce

Building the docker image for alas will reproduce the issue.

### Expected behavior

The installation should succeed.

### Relevant log output

```Text
PS D:\AppFiles\AzurLaneAutoScript\docker-instance> docker compose up
time="2026-09-14T02:24:46-04:00" level=warning msg="D:\\AppFiles\\AzurLaneAutoScript\\docker-instance\\docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion"
[+] up 1/1
 ! Image a... pull access denied for alas-bullseye, repository does not exist or may require 'docker login': denied: requested access to the resource is denied 0.4s
#1 [internal] load local bake definitions
#1 reading from stdin 592B done
#1 DONE 0.0s

#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 932B 0.0s done
#2 DONE 0.0s

#3 [internal] load metadata for docker.io/library/python:3.7-slim-bullseye
#3 DONE 0.2s

#4 [internal] load .dockerignore
#4 transferring context: 80B done
#4 DONE 0.0s

#5 [1/4] FROM docker.io/library/python:3.7-slim-bullseye@sha256:603879aeff06df23f0d392a19f7ce02aa85e14e6b4d25fd9647b116e0cb93467
#5 DONE 0.0s

#6 [internal] load build context
#6 transferring context: 38B done
#6 DONE 0.0s

#7 [2/4] WORKDIR /app/AzurLaneAutoScript
#7 CACHED

#8 [3/4] COPY requirements.txt /tmp/requirements.txt
#8 CACHED

[...]

PR #5986, opened 2026-09-14
Fix: Debian EOL causing docker installation failed

fybmain:
A mitigation of the issue #5985.

Note: The regular security support for Debian 12 (Bookworm) has also ended on since July 11, 2026. It has already been an obsolete version. However, the attempt to upgrade to the latest Debian version 13 (Trixie) will fail due to the dependecy conflict of PyAV. So the PR should be viewed as a mitigation rather than a long-term solution.


fybmain:
Tested the built image with my AzurLane game.
Tested features: Farm/Main Event/Event EventDaily/EventDailyA EventDaily/EventDailyB EventDaily/EventDailySP Reward/Commission Reward/Tactical Reward/Research Reward/Meowfficer Reward/Reward DailyMission/Daily DailyMission/Hard DailyMission/Exercise DailyMission/ShopFrequent DailyMission/ShopOnce DailyMission/Shipyard DailyMission/Freebies DailyMission/Minigame OpSi/METAAssists
All functioning correctly.

PR #6001, opened 2026-09-20
Fix 16-4 boss spawn after the fifth battle

954708775:
## Summary

- restore the normal-mode 16-4 boss spawn to battle 4 (the fifth combat)
- prevent ALAS from clearing three unnecessary mobs after the boss is already visible
- leave clear-mode behavior unchanged

## Evidence

A normal-mode runtime log shows a BO tile at J1 immediately after BATTLE_4. With the current spawn table expecting the boss at battle 8, ALAS proceeds to BATTLE_5 and clears C3 instead.

The initial W16 implementation also used battle 4 for the boss. Commit b993a5e8 changed only the normal spawn table to battle 8, while spawn_data_loop remained at battle 4.

## Verification

- python3 -m py_compile campaign/campaign_main/campaign_16_4.py
- imported MAP and asserted both spawn tables end with {battle: 4, boss: 1}
- git diff --check
- independent diff and runtime-log review

Not tested: full live 16-4 campaign run.


campaign/campaign_main/campaign_16_4.py
guoh064, line 152: 这里的现有部分不能处理boss吗？运行有问题的话可以发issue，附上log。

guoh064 (changes requested):
先看一下原来代码在干什么和原来运行log是什么吧
~~上来就大刀阔斧是不是有点奇怪~~

campaign/campaign_main/campaign_16_4.py
954708775: 现有这部分可以处理 Boss，问题出在它执行前的地图扫描时序。第五场结束后的 full scan 按当前 spawn_data 预期 BO:0，日志显示 `[enemy_missing] ... BO: 0` 后立即 `All spawn found, Early stopped`，所以 `self.map.select(is_boss=True)` 此时拿不到 Boss。脚本先选定并开始前往 C3，镜头途中经过 J1 才识别到 `BO`；C3 战斗结束后又按 BO:0 提前停止，随后继续清 A2。完整关键日志和复现条件已整理到 #6002。先不继续改 PR，等确认修复点应放在 spawn_data 还是扫描逻辑。

